PR #10, opened 2021-09-09
Dcs 1103 show incoming moves

louisenorris:


GurnankCheema (approved):
LGTM

integration_tests/plugins/index.ts
```diff
@@ -16,6 +16,6 @@ export default (on: (string, Record) => void): void => {
 
     stubTokenVerificationPing: tokenVerification.stubPing,
 
-    stubIncomingMovements: welcome.stubIncomingMovements,
+    stubIncomingMovements: activeCaseLoadId => welcome.stubIncomingMovements(activeCaseLoadId),
```
andrewrlee, line 19: meh.. don't this needed changing. This was assigning a function to a function so shouldn't have needed to know about changes to how the function was being called

server/routes/choosePrisonerController.test.ts
```diff
@@ -54,28 +67,35 @@ describe('Choose Prisoner Controller', () => {
...
-    it('should return incoming movements', async () => {
-      await controller.view()(req, res, null)
+afterEach(() => {
+  jest.resetAllMocks()
+})
 
-      expect(res.render).toHaveBeenCalledWith('pages/choosePrisoner.njk', {
-        incomingMovements,
+describe('GET /confirm-arrival/choose-prisoner', () => {
+  it('should contain alphabetically sorted incoming movements grouped by type', () => {
```
andrewrlee, line 83: sorting is an implementation detail that you don't need to expose at this level - the test name suggests that this is a feature of the module under test

server/routes/choosePrisonerController.ts
```diff
@@ -6,7 +6,8 @@ export default class ChoosePrisonerController {
 
   public view(): RequestHandler {
     return async (req, res) => {
-      const incomingMovements = await this.incomingMovementsService.getIncomingMovements('MDI')
+      const { activeCaseLoadId } = res.locals.user
+      const incomingMovements = await this.incomingMovementsService.groupByMoveType(activeCaseLoadId)
```
andrewrlee, line 10: `getMovesGroupedByType` ??

server/services/incomingMovementsService.ts
```diff
@@ -9,9 +10,29 @@ export default class IncomingMovementsService {
     private readonly welcomeApiFactory: (token: string) => WelcomeApi
   ) {}
 
+  private moveTypeGroupingFunction(item: Movement): string {
```
andrewrlee, line 13: rather than string might be nicer to have an enum to represent this set of closed values

Also might be nice the represent the null explicitly - either as part of the return value or even an explicit enum value `OTHER` or whatever

(note upper case to represent that these are constants)

server/services/incomingMovementsService.ts
```diff
@@ -9,9 +10,29 @@ export default class IncomingMovementsService {
...
+  }
+
+  private sortAlphabetically(movements: Movement[]): Movement[] {
+    return movements.sort((a, b) => {
+      const result = a.lastName.localeCompare(b.lastName)
+      return result !== 0 ? result : a.firstName.localeCompare(b.firstName)
+    })
+  }
+
   public async getIncomingMovements(agencyId: string): Promise<Movement[]> {
```
andrewrlee, line 27: guess this can be private?

server/services/incomingMovementsService.ts
```diff
@@ -9,9 +10,29 @@ export default class IncomingMovementsService {
...
     const today = moment.now().toString()
     const token = await this.hmppsAuthClient.getSystemClientToken()
-    return this.welcomeApiFactory(token).getIncomingMovements(agencyId, today)
+    const movements = await this.welcomeApiFactory(token).getIncomingMovements(agencyId, today)
+    return this.sortAlphabetically(movements)
+  }
+
+  public async groupByMoveType(agencyId: string): Promise<Map<string, Movement[]>> {
+    const movements = await this.getIncomingMovements(agencyId)
+    return groupBy(movements, (movement: Movement) => this.moveTypeGroupingFunction(movement))
```
andrewrlee, line 36: very meh - 

if you look at the function `moveTypeGroupingFunction` , what it does and its signature, it doesn't need to be used as a grouping function - could be called something like `getMoveType`, `calculateMoveType`, `determineMoveType`... etc

Makes it a bit easier to reuse.

And super meh.. think you might be able to call this like: 
`groupBy(movements, this.moveTypeGroupingFunction)`
(possibly)

server/services/incomingMovementsService.ts
andrewrlee: (Using an explicit value rather than `null` is an example of the [https://en.wikipedia.org/wiki/Null_object_pattern](Null object pattern))

integration_tests/plugins/index.ts
louisenorris: reverted back

server/routes/choosePrisonerController.test.ts
louisenorris: Updated the test name to 'should render /confirm-arrival/choose-prisoner page'

server/routes/choosePrisonerController.ts
louisenorris: updated function name to suggested `getMovesForToday`

server/services/incomingMovementsService.ts
louisenorris: Yep, made private as suggested.

server/services/incomingMovementsService.ts
louisenorris: grouping function renamed but wasn't able to call it as just `this.moveTypeGroupingFunction`

PR #21, opened 2021-09-24
DCS-1187 displays prisoner's actual image or a placeholder image

GurnankCheema:
A prisoner's image, or a placeholder if image not available, will be displayed. 

Screenshot 2021-09-24 at 11 13 44


andrewrlee (commented):
LGTM but needs some level of testing

Could add cypress test to return an image from wiremock and potentially verify it's loaded when accessing content

server/routes/index.ts
```diff
@@ -19,5 +20,8 @@ export default function routes(services: Services): Router {
   const temporaryAbsencesController = new TemporaryAbsencesController(services.temporaryAbsencesService)
   get('/confirm-arrival/return-from-temporary-absence', temporaryAbsencesController.view())
 
+  const imageController = new ImageController(services.incomingMovementsService)
```
andrewrlee, line 23: could maybe call it prisoner or offender controller? 

Image is quite generic to be sitting on such a specific url

GurnankCheema: renamed to prisonerController

connormaglynn (approved):
LGTM 😃

PR #28, opened 2021-10-07
DCS-1204 Unmatched court arrivals identified

GurnankCheema:
The image and prisoner name of unmatched court arrivals are clickable and take user to the Confirm arrival page

Screenshot 2021-10-07 at 16 34 14

### DIFF
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -25,7 +25,7 @@ export default function routes(services: Services): Router {
   get('/prisoner/:prisonNumber/image', prisonerController.getImage())
 
   const confirmArrivalController = new ConfirmArrivalController()
-  get('/prisoner/:UUID/confirm-arrival', confirmArrivalController.confirmArrival())
+  get('/prisoners/:UUID/confirm-arrival', confirmArrivalController.confirmArrival())
 
   return router
 }
--- a/server/views/partials/expectedArrivalCardMacro.njk
+++ b/server/views/partials/expectedArrivalCardMacro.njk
@@ -17,7 +17,7 @@
                 <header class="app-card__header govuk-!-margin-bottom-4">
                     <h2 class="app-card__title" data-qa="{{moveType}}-title-{{loopIndex}}">
                         {% if arrival.fromLocationType === 'COURT' and not arrival.prisonNumber %}
-                            <a href="/prisoner/{{ arrival.id }}/confirm-arrival" class="govuk-link govuk-link--no-visited-state" data-qa="confirm-arrival"> {{ arrival.lastName }}, {{ arrival.firstName }}</a>
+                            <a href="/prisoners/{{ arrival.id }}/confirm-arrival" class="govuk-link govuk-link--no-visited-state" data-qa="confirm-arrival"> {{ arrival.lastName }}, {{ arrival.firstName }}</a>
                         {% else %}
                             {{ arrival.lastName }}, {{ arrival.firstName }}
                         {%  endif %}
--- a/integration_tests/pages/confirmArrival.ts
+++ b/integration_tests/pages/confirmArrival.ts
@@ -6,7 +6,7 @@ export default class ConfirmArrivalPage extends Page {
   }
 
   static goTo(): ConfirmArrivalPage {
-    cy.visit('/prisoner/12345/confirm-arrival')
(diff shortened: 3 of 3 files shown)


integration_tests/integration/choosePrisoner.spec.ts
```diff
@@ -89,4 +90,19 @@ context('SignIn', () => {
         expect(altText).equal('Headshot of Stanton, Harry')
       })
   })
+
+  it('Only the names and images of prisoners arriving from court and without a prisoner number will be clickable', () => {
+    cy.signIn()
+    const choosePrisonerPage = Page.verifyOnPage(ChoosePrisonerPage)
+
+    choosePrisonerPage.prisonerImageLink().should('have.length', 1)
+    choosePrisonerPage.prisonerImageLink().eq(0).click()
```
andrewrlee, line 99: This behaviour doesn't match either the ticket description or the prototype?

integration_tests/integration/choosePrisoner.spec.ts
```diff
@@ -89,4 +90,19 @@ context('SignIn', () => {
...
+    Page.verifyOnPage(ConfirmArrivalPage)
+
+    cy.go('back')
+
+    choosePrisonerPage.prisonerNameLink().should('have.length', 1)
```
andrewrlee, line 104: meh.. by specifying `prisonerNameLink` we're binding to the specific structure of the page. The tests don't really need to know that the action is associated with where the prisoners name is or that it is a link. 

We can avoid this by jumping to a higher layer of abstraction and call it `confirmArrival`

The benefit of this is that the tests are more decoupled so less susceptible from changes in design.

We could represent this control as a massive button to the right of the row and not update the tests.

server/routes/index.ts
```diff
@@ -22,6 +22,7 @@ export default function routes(services: Services): Router {
 
   const prisonerController = new PrisonerController(services.expectedArrivalsService)
   get('/prisoner/:prisonNumber/image', prisonerController.getImage())
+  get('/prisoner/:UUID/confirm-arrival', prisonerController.getPrisoner())
```
andrewrlee, line 25: can we call this `prisonerController.confimArrival` to match the name of the page?

It could probably do with being its own controller.. but happy to leave for now.

We need to keep an eye on this files as its going to get messy really quickly

server/views/partials/expectedArrivalCardMacro.njk
```diff
@@ -1,22 +1,38 @@
 {% macro expectedArrivalCardMacro(arrival, moveType, loopIndex) %}
+    {% if arrival.fromLocationType === 'COURT' and not arrival.prisonNumber %}
+        {% set isUnmatchedCourtArrival = true %}
+    {% endif %}
```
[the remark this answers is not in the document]
GurnankCheema: I removed it and put it straight in 'if' block as it's only used once

server/views/partials/expectedArrivalCardMacro.njk
```diff
@@ -1,22 +1,38 @@
...
                 <aside class="app-card__aside">
-                    <img
-                    {% if arrival.prisonNumber  %}
-                        src= "/prisoner/{{arrival.prisonNumber}}/image"
-                    {% else %}
-                        src= "/assets/images/placeholder-image.png"
+                    {% if isUnmatchedCourtArrival %}
+                      <a href="/prisoner/{{ arrival.id }}/confirm-arrival" class="app-card__image" data-qa="prisoner-image-link">
+                    {% endif %}
+                        <img
```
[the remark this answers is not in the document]
GurnankCheema: if block no longer needed

server/routes/index.ts
GurnankCheema: Name changed and moved out into its own controller

integration_tests/integration/choosePrisoner.spec.ts
```diff
@@ -89,4 +90,19 @@ context('SignIn', () => {
         expect(altText).equal('Headshot of Stanton, Harry')
       })
   })
+
+  it('Only the names and images of prisoners arriving from court and without a prisoner number will be clickable', () => {
```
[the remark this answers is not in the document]
GurnankCheema: Updated as per discussion

PR #29, opened 2021-10-11
DCS-1214 css image fix for temporary absences

louisenorris:


integration_tests/integration/temporaryAbsences.spec.ts
```diff
@@ -15,4 +16,46 @@ context('A user can view all current temporary absences', () => {
...
+    temporaryAbsencesPage
+      .prisonerImage(0)
+      .should('have.attr', 'alt')
+      .then(altText => {
+        expect(altText).equal('Headshot of Doe, John')
+      })
+  })
+
+  it('A user will see placeholder image as prisoner has no image', () => {
+    cy.task('stubPrisonerImage', { prisonerNumber: 'G0015GD', imageFile: '/placeholder-image.png' })
```
andrewrlee, line 42: Think you can ditch this test - the assertions are the same

To test this properly, you wouldn't want to stubPrisonerImage - this says that the prisoner _does_ have an image (its as if the placeholder had been associated with the prisoner in nomis). 

So you'd want a 404 to be sent from the backend (`stubMissingPrisonerImage`)
And then you'd need to actually test the content of the image matches the placeholder image. 

It would be nice to have but possibly too much work

andrewrlee (approved):
LGTM

PR #41, opened 2021-11-01
✨DCS-1260: Add confirm link to Matched COURT Arrivals that are not currently in prison

connormaglynn:


integration_tests/mockApis/responses/expectedArrivals.ts
```diff
@@ -69,6 +71,7 @@ export default [
     date: '2021-09-01',
     fromLocation: 'Reading',
     fromLocationType: 'COURT',
+    isCurrentPrisoner: true,
```
connormaglynn, line 74: Only COURT scenario left that doesn't show a link

server/views/partials/expectedArrivalCardMacro.njk
```diff
@@ -12,7 +12,7 @@
             <div class="app-card__content">
                 <header class="app-card__header govuk-!-margin-bottom-4">
                     <h2 class="app-card__title" data-qa="{{moveType}}-title-{{loopIndex}}">
-                        {% if arrival.fromLocationType === 'COURT' and not arrival.prisonNumber %}
+                        {% if arrival.fromLocationType === 'COURT' and not arrival.isCurrentPrisoner %}
```
connormaglynn, line 15: So from court we have:

- Matched & out of prison = Has prisonNumber and&isCurrentPrisoner is false = show link
- Matched & In prison = Has prisonNumber & isCurrentPrisoner  is true = no link
- Unmatched = No prisonNumber &  isCurrentPrisoner is false = show link

think the above logic covers all these three scenarios

bell-pa (approved):
LGTM

PR #42, opened 2021-11-02
DCS-1232 added universal analytics event to capture data around create offender record and booking

louisenorris:


connormaglynn (approved):
lgtm

PR #43, opened 2021-11-04
🔐DCS-1164: restrict Confirm Arrival and Temporary Absence journey to reception users

connormaglynn:


server/authentication/role.ts
```diff
@@ -0,0 +1,5 @@
+const enum Role {
```
connormaglynn, line 1: Did some reading around this, apparently not good practice to export **const enum**s in Typescript. Article here: https://ncjamieson.com/dont-export-const-enums/

Not too great at my Typescript to know whether we care about this. But I thought I'd point it out in case anyone had an alternative?

Also not too happy with the location of this file, might be better suited in a /types folder? Any thoughts?

server/routes/index.ts
```diff
@@ -5,14 +5,18 @@ import ConfirmArrivalController from './confirmArrivalController'
...
 import { Services } from '../services'
 import TemporaryAbsencesController from './temporaryAbsencesController'
+import Role from '../authentication/role'
 
 export default function routes(services: Services): Router {
   const router = express.Router()
-  const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
-  const post = (path: string, handler: RequestHandler) => router.post(path, asyncMiddleware(handler))
+  const get = (path: string, handler: RequestHandler, authorisedRoles?: Role[]) =>
+    router.get(path, authorisationForUrlMiddleware(authorisedRoles), asyncMiddleware(handler))
```
connormaglynn, line 17: I thought it made sense to have authorisation middleware specific for the URL. Makes it clear who can access what routes. Not seen this pattern in other services so happy for criticisms on this 👍

server/middleware/authorisationMiddleware.ts
```diff
@@ -13,6 +14,11 @@ export default function authorisationMiddleware(authorisedRoles: string[] = []):
         return res.redirect('/authError')
       }
 
+      res.locals.user = {
+        roles,
+        isReceptionUser: roles.includes(Role.RECEPTION_USER),
```
connormaglynn, line 19: Kind of just plopped this bit of code in this middleware as it's relevant to authorisation and I didn't think it was big enough to create it's own middleware. Maybe if it becomes bigger (more complex access control based on roles) - we can separate into a **buildApplicationRoles** middleware or something along those lines

server/middleware/authorisationForUrlMiddleware.ts
```diff
@@ -0,0 +1,14 @@
+import { RequestHandler } from 'express'
+
+import logger from '../../logger'
+import Role from '../authentication/role'
+
+export default function authorisationForUrlMiddleware(authorisedRoles: Role[] = []): RequestHandler {
+  return (req, res, next) => {
+    if (authorisedRoles.length && !res.locals.user.roles.some((role: Role) => authorisedRoles.includes(role))) {
```
connormaglynn, line 8: **authorisedRoles.length** says if no roles are passed, allow anyone to see the route. This is the desired behaviour

bell-pa (approved):
LGTM

PR #45, opened 2021-11-05
DCS-1255: move to do prison lookup using prison registry

bell-pa:
longDescription changed to description (refer to the story for detail)


louisenorris (approved):
LGTM

PR #60, opened 2021-11-22
DCS-1277  get individual transfer from welcome-api

GurnankCheema:
all transfer calls will eventually be moved into transferService


server/services/transfersService.ts
```diff
@@ -0,0 +1,23 @@
+import type { Transfer } from 'welcome'
+import type { RestClientBuilder, WelcomeClient, HmppsAuthClient } from '../data'
+
+export enum LocationType {
```
andrewrlee, line 4: not needed?

GurnankCheema: no

andrewrlee (approved):
LGTM

PR #63, opened 2021-11-23
DCS-1303 add location banner

louisenorris:


server/middleware/setUpCurrentUser.ts
```diff
@@ -11,7 +11,7 @@ export default function setUpCurrentUser(services: Services): Router {
   const router = Router({ mergeParams: true })
 
   router.use(auth.authenticationMiddleware(tokenVerifier))
-  router.use(populateCurrentUser(services.userService))
+  router.use(populateCurrentUser(services.userService, services.expectedArrivalsService))
```
andrewrlee, line 14: umm.. this seems a bit off - populateCurrentUser is domain agnostic and we're pushing in the thing responsible for a chunk of our domain.

The only reason we need this is for: 
`expectedArrivalsServices.getPrison(user.activeCaseLoadId)`

the concept of `getPrison` doesn't seem very related to conceptual area of expected arrivals.

Maybe we should extract it into its own service

louisenorris: Sorted

andrewrlee (dismissed):
LGTM

andrewrlee (dismissed):
LGTM

server/views/partials/header.njk
```diff
@@ -245,3 +246,18 @@
...
+        {% if phaseName %}
+            {{ govukTag({
+                text: phaseName,
+                classes: 'govuk-!-margin-right-1 ' + phaseNameColour
+            }) }}
+        {% endif %}
+        <span class="gtm-active-case-load" data-qa="active-case-load">{{ user.activeCaseLoad.description }}</span>
+    </span>
+    {% if user.userCaseLoads.length > 1 %}
+      <a class="location-bar__link" href="https://digital-dev.prison.service.justice.gov.uk/change-caseload" data-qa="change-location-link">Change your location</a>
```
andrewrlee, line 261: this needs to be parameterised

server/config.ts
```diff
@@ -35,6 +35,7 @@ export default {
   https: production,
   staticResourceCacheDuration: 20,
   phaseName: process.env.SYSTEM_PHASE,
+  dpsUrl: get('DPS_URL', 'http://localhost:3000', requiredInProduction),
```
louisenorris, line 38: Not sure is localhost on 300 is correct here

server/views/partials/header.njk
louisenorris: Nope, I've added one though

andrewrlee (approved):
LGTM

PR #66, opened 2021-11-26
DCS-1280 add controller, service and view for prison transfer

GurnankCheema:
Integration test to be added in another ticket

view looks like this

transfer added to prison roll 26:11:21

### DIFF
--- a/server/routes/checkTransferController.ts
+++ b/server/routes/checkTransferController.ts
@@ -28,12 +28,10 @@ export default class CheckTransferController {
         req.hostname
       )
 
-      req.flash('prisoner', [
-        {
-          firstName: data.firstName,
-          lastName: data.lastName,
-        },
-      ])
+      req.flash('prisoner', {
+        firstName: data.firstName,
+        lastName: data.lastName,
+      })
 
       await this.transfersService.confirmTransfer(username, prisonNumber)
       res.redirect(`/prisoners/${prisonNumber}/confirm-transfer`)
--- a/server/routes/checkTransferController.ts
+++ b/server/routes/checkTransferController.ts
@@ -1,5 +1,6 @@
 import { RequestHandler } from 'express'
 import TransfersService from '../services/transfersService'
+import raiseAnalyticsEvent from '../raiseAnalyticsEvent'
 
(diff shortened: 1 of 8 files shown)


server/routes/confirmTransferAddedToRollController.ts
```diff
@@ -0,0 +1,30 @@
...
+    private readonly transfersService: TransfersService
+  ) {}
+
+  public view(): RequestHandler {
+    return async (req, res) => {
+      const { prisonNumber } = req.params
+      const { activeCaseLoadId } = res.locals.user
+      const { firstName, lastName } = req.flash('prisoner')?.[0] as Record<string, string>
+
+      if (!firstName && !lastName) {
```
GurnankCheema, line 17: there are people who only have a one name, i.e. just a firstName!

server/routes/confirmTransferAddedToRollController.ts
```diff
@@ -0,0 +1,30 @@
...
+        return res.redirect('/confirm-arrival/choose-prisoner')
+      }
```
GurnankCheema, line 19: there are people who only have just one name i.e.a firstName!

server/routes/checkTransferController.ts
```diff
@@ -16,8 +17,24 @@ export default class CheckTransferController {
...
+        `AgencyId: ${activeCaseLoadId}, From: ${data.fromLocation}, Type: 'PRISON',`,
+        req.hostname
+      )
+
+      req.flash('prisoner', {
+        firstName: data.firstName,
+        lastName: data.lastName,
+      })
+
+      await this.transfersService.confirmTransfer(username, prisonNumber)
```
andrewrlee, line 36: maybe we should swap round the raiseAnalyticsEvent so it only happens in the confirmTransfer was successful?

andrewrlee: we should probably do the same for confirmArrival also if its been done that way

GurnankCheema: ok

andrewrlee (approved):
LGTM

PR #69, opened 2021-12-01
DCS-1306 added back to dps links

louisenorris:


integration_tests/integration/confirmTransferAddedToRoll.spec.ts
```diff
@@ -34,4 +34,17 @@ context('Confirm transfer added To roll', () => {
...
+  it('Should display Back to Digital Prisons Services link with the correct href', () => {
+    const checkTransferPage = CheckTransferPage.goTo(expectedArrivals.prisonTransfer.prisonNumber)
+    checkTransferPage.addToRoll().click()
+    const confirmTransferAddedToRollPage = Page.verifyOnPage(ConfirmTransferAddedToRollPage)
+    confirmTransferAddedToRollPage
+      .backToDigitalPrisonServices()
+      .should('contain', 'Back to Digital Prison Services')
+      .should('have.attr', 'href')
+      .then(href => {
+        expect(href).to.equal('https://digital-dev.prison.service.justice.gov.uk')
```
louisenorris, line 47: Don't think this url needs to be a parameter if it's external from the service in these tests but might be wrong. Looking at what we did with the feedback link url we just included it in the test like above: https://github.com/ministryofjustice/hmpps-book-video-link/blob/6e0a1f25a71e143780c66909a017c2a9348ee52d/integration-tests/integration/homePage.spec.js#L80

andrewrlee: Yes, this is fine - you've set it to this in feature.env so it's fine to hard code it to the same value

andrewrlee (approved):
LGTM

PR #88, opened 2022-01-06
DCS-1261 handle errors from welcome-api

GurnankCheema:


integration_tests/integration/featureNotAvailable.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    imprisonmentStatusPage.imprisonmentStatusRadioButton('on-remand').click()
+    imprisonmentStatusPage.continue().click()
+
+    const checkAnswersPage = CheckAnswersPage.goTo(expectedArrival.id)
+
+    checkAnswersPage.addToRoll().click()
+    Page.verifyOnPage(FeatureNotAvailable)
+  })
+
+  it('Should not display feature-not-available page', () => {
```
andrewrlee, line 43: I don't think we need this test as its behaviour is covered in other places

integration_tests/integration/featureNotAvailable.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    sexPage.sexRadioButtons('M').click()
+    sexPage.continue().click()
+
+    const imprisonmentStatusPage = ImprisonmentStatusPage.goTo(expectedArrival.id)
+    imprisonmentStatusPage.imprisonmentStatusRadioButton('on-remand').click()
+    imprisonmentStatusPage.continue().click()
+
+    const checkAnswersPage = CheckAnswersPage.goTo(expectedArrival.id)
+
+    checkAnswersPage.addToRoll().click()
```
andrewrlee, line 39: could we be a bit more explicit here? Presumably it fails as we haven't stubbed out the response to create booking and wiremock returns a 404.

Could we show that stubbing this call to produce a 400 response produces this scenario?

integration_tests/integration/featureNotAvailable.spec.ts
GurnankCheema: Yes, I was in two minds about this one.

server/data/welcomeClient.test.ts
```diff
@@ -157,6 +158,40 @@ describe('welcomeClient', () => {
...
+      fakeWelcomeApi
+        .post(`/arrivals/${id}/confirm`, newOffender)
+        .matchHeader('authorization', `Bearer ${token}`)
+        .reply(400)
+
+      const output = await welcomeClient.createOffenderRecordAndBooking(id, newOffender)
+      return expect(output).toBe(null)
+    })
+
+    it('server error', async () => {
```
andrewrlee, line 171: we can get rid of this test now?

GurnankCheema: removed

integration_tests/integration/featureNotAvailable.spec.ts
GurnankCheema: done

andrewrlee (approved):
LGTM

PR #89, opened 2022-01-07
DCS-1379 updated returning from temporary absence url

louisenorris:
Note: Contents of confirmTemporaryAbsenceAddedToRoll.njk view page to be added in a future ticket, draft view page included as part of this PR to play out navigation from temporary absence check page.

### DIFF
--- a/server/services/temporaryAbsencesService.ts
+++ b/server/services/temporaryAbsencesService.ts
@@ -1,6 +1,6 @@
 import type { TemporaryAbsence } from 'welcome'
 import type { RestClientBuilder, WelcomeClient, HmppsAuthClient } from '../data'
-import { compareByFullName, convertToTitleCase } from '../utils/utils'
+import { compareByFullName } from '../utils/utils'
 
 export default class TemporaryAbsencesService {
   constructor(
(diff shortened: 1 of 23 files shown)


server/utils/nunjucksSetup.ts
```diff
@@ -48,6 +49,10 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
     return `${array[0][0]}. ${array.reverse()[0]}`
   })
 
+  njkEnv.addFilter('formatName', (name: string) => {
+    return convertToTitleCase(name)
+  })
```
andrewrlee, line 54: we tend to handle this on the backend for arrivals - we should probably do the same for temp absences

server/services/temporaryAbsencesService.ts
```diff
@@ -13,4 +13,14 @@ export default class TemporaryAbsencesService {
     const temporaryAbsences = await this.welcomeClientFactory(token).getTemporaryAbsences(agencyId)
     return temporaryAbsences.sort(compareByFullName)
   }
+
+  public async getTemporaryAbsence(agencyId: string, prisonNumber: string): Promise<TemporaryAbsence> {
+    const token = await this.hmppsAuthClient.getSystemClientToken()
+    const temporaryAbsence = await this.welcomeClientFactory(token).getTemporaryAbsence(agencyId, prisonNumber)
+    return {
+      ...temporaryAbsence,
+      firstName: convertToTitleCase(temporaryAbsence.firstName),
```
andrewrlee, line 22: formatting twice?

louisenorris: Agree it isn't great - seemed less code than formatting in both the service get methods but I guess I could have extracted a private formatting method there... will sort in the backend instead as suggested - thank you

louisenorris: Removed on latest commit

server/utils/nunjucksSetup.ts
louisenorris: Removed on latest commit

andrewrlee (approved):
LGTM

PR #91, opened 2022-01-12
DCS-1397 Restructuring routes so matches structure of the application

andrewrlee:
Moved routes and views into a structure that more represents the structure of the application as discussed in our byte sized architectural sessions.

We previously had a very flat structure with one routes file and all the route modules in one directory.

Now the folder structure has this shape:

```
routes/
    bookedtoday/
        arrivals/
        transfers/
    temporaryabsences/
``` 

Each folder has an index file that exposes a `Router` and these are all wired up together in a central `/routes/index.ts` file.

There's a couple of anonamalies:
* `Prisoner controller` only provides images of prisoners which is used across the app so sits under `routes/`
* `Select prisoner controller` is the entrypoint for both `arrivals` and `transfers` so sits in `bookedtoday`


server/index.ts
```diff
@@ -1,6 +1,6 @@
 import createApp from './app'
 import { services } from './services'
 
-const app = createApp(services)
+const app = createApp(services())
```
andrewrlee, line 4: we had a problem where importing services anywhere would cause the redis client to make a connection.
This would prevent tests from exiting properly so made this lazy so the connection is only called when the services function is actually called.

This allows us to import Service without creating connections

server/middleware/caseloadCheckMiddleware.test.ts
```diff
@@ -1,5 +1,5 @@
 import caseloadCheck from './caseloadCheckMiddleware'
-import { mockNext, mockRequest, mockResponse } from '../routes/testutils/requestTestUtils'
+import { mockNext, mockRequest, mockResponse } from '../routes/__testutils/requestTestUtils'
```
andrewrlee, line 2: moved test utils into a sub directory to make it distinct compared to the other folders that represent different parts of the app

server/middleware/setUpAuthentication.ts
```diff
@@ -16,7 +16,7 @@ export default function setUpAuth(): Router {
 
   router.get('/autherror', (req, res) => {
     res.status(401)
-    return res.render('autherror')
+    return res.render('pages/autherror.njk')
```
andrewrlee, line 19: previously we had some pages outside of the pages folder - I've made sure this no longer the case

server/routes/bookedtoday/arrivals/confirmAddedToRollController.ts
```diff
@@ -20,7 +20,7 @@ export default class ConfirmAddedToRollController {
       const prison = await this.prisonService.getPrison(activeCaseLoadId)
       clearImprisonmentStatus(res)
       clearSex(res)
-      return res.render('pages/confirmAddedToRoll', { data, prison, offenderNumber })
+      return res.render('pages/bookedtoday/arrivals/confirmAddedToRoll.njk', { data, prison, offenderNumber })
```
andrewrlee, line 23: some places we have the `njk` suffix and others we don't - its not necessary so we could remove it everywhere

server/services/index.ts
```diff
@@ -6,24 +6,27 @@ import ImprisonmentStatusesService from './imprisonmentStatusesService'
...
 
-const { hmppsAuthClient, welcomeClientBuilder } = dataAccess
-const userService = new UserService(hmppsAuthClient, welcomeClientBuilder)
-const expectedArrivalsService = new ExpectedArrivalsService(hmppsAuthClient, welcomeClientBuilder)
-const temporaryAbsencesService = new TemporaryAbsencesService(hmppsAuthClient, welcomeClientBuilder)
-const imprisonmentStatusesService = new ImprisonmentStatusesService(hmppsAuthClient, welcomeClientBuilder)
-const transfersService = new TransfersService(hmppsAuthClient, welcomeClientBuilder)
-const prisonService = new PrisonService(hmppsAuthClient, welcomeClientBuilder)
+export const services = () => {
+  const { hmppsAuthClient, welcomeClientBuilder } = dataAccess()
```
andrewrlee, line 10: only call `dataAccess` to create connection if we are calling the `services` function

louisenorris (approved):
LGTM

PR #92, opened 2022-01-13
DCS-1286 added client layer for confirmTemporaryAbsence and typecheck fixes

louisenorris:


integration_tests/index.d.ts
```diff
@@ -3,6 +3,7 @@ declare namespace Cypress {
     /**   * Custom command to signIn. Set failOnStatusCode to false if you expect and non 200 return code
      * @example cy.signIn({ failOnStatusCode: boolean })
      */
-    signIn<S = unknown>(options?: { failOnStatusCode: false }): Chainable<S>
+
+    signIn<S = unknown>(options?: { failOnStatusCode: false | true }): Chainable<S>
```
louisenorris, line 7: Typecheck failing with type false set as we change this to true elsewhere (in integration_tests/support/commands.ts Not sure if this is the correct fix.

server/routes/bookedtoday/index.ts
```diff
@@ -1,4 +1,4 @@
-import express, { type Router } from 'express'
+import express, { Router } from 'express'
```
louisenorris, line 1: Removed type as caused typecheck to fail in multiple places. Not sure if this is the correct fix.

integration_tests/index.d.ts
louisenorris: This has been fixed in a separate ticket

server/routes/bookedtoday/index.ts
andrewrlee: Don't think this is right - it doesn't fail for me. Did you definitely run npm install?

Support for adding `type` to individually named imports in the [latest version of typescript](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names)

server/routes/temporaryabsences/checkTemporaryAbsenceController.test.ts
```diff
@@ -3,13 +3,17 @@ import request from 'supertest'
...
 
 import Role from '../../authentication/role'
 import config from '../../config'
 
 jest.mock('../../services/temporaryAbsencesService')
 const temporaryAbsencesService = new TemporaryAbsencesService(null, null) as jest.Mocked<TemporaryAbsencesService>
 let app: Express
+const flash = jest.fn()
+
+jest.mock('../../raiseAnalyticsEvent')
```
andrewrlee, line 16: this direct import of raiseAnalyticsEvent doesn't really fit the rest of the way the app is structured - I think this was copied in from DPS core.

We'd usually inject in our dependencies, it would also be nice to be able to switch it off - don't really want it to be trying to fire in integration tests for instance. 

Maybe we should add a ticket to change this to a more conventional dependency - which doesn't do anything when disabled.

server/data/welcomeClient.test.ts
```diff
@@ -135,6 +136,23 @@ describe('welcomeClient', () => {
     })
   })
 
+  describe('confirmTemporaryAbsence', () => {
+    const prisonNumber = 'A1234AB'
+    const agencyId: AgencyId = {
```
michalnawrockidj, line 141: The request object is more complex where AgancyId is one of the parameters maybe will be better to rename it to ConfirmTemporaryAbsenceRequest.

server/routes/temporaryabsences/checkTemporaryAbsenceController.ts
```diff
@@ -16,6 +17,25 @@ export default class CheckTemporaryAbsenceController {
   public addToRoll(): RequestHandler {
     return async (req, res) => {
       const { prisonNumber } = req.params
+      const { username } = req.user
+      const { activeCaseLoadId } = res.locals.user
+      const data = await this.temporaryAbsencesService.getTemporaryAbsence(activeCaseLoadId, prisonNumber)
+
+      await this.temporaryAbsencesService.confirmTemporaryAbsence(username, prisonNumber, {
+        agencyId: activeCaseLoadId,
```
andrewrlee, line 25: I think it would be better to just pass in the `activeCaseLoadId` directly rather than this `AgencyId` Type. 

This leaks client implementation specific types up through the service and web layer - your controller shouldn't need to transform its data to a structure that is only required by the client if it doesn't have to.  

You could do the same in the service too - pass the agencyId straight through and build the structure solely in the client?

server/routes/temporaryabsences/confirmTemporaryAbsenceAddedToRollController.ts
```diff
@@ -1,9 +1,27 @@
...
+  public constructor(private readonly prisonService: PrisonService) {}
+
   public view(): RequestHandler {
     return async (req, res) => {
-      return res.render('pages/temporaryabsences/confirmTemporaryAbsenceAddedToRoll.njk', {})
+      const { prisonNumber } = req.params
+      const { activeCaseLoadId } = res.locals.user
+      const { firstName, lastName } = req.flash('prisoner')?.[0] as Record<string, string>
+
+      if (!firstName && !lastName) {
```
andrewrlee, line 13: wouldn't you want to redirect if either of these weren't present?

server/views/pages/temporaryabsences/confirmTemporaryAbsenceAddedToRoll.njk
```diff
@@ -1 +1,37 @@
...
+{% set pageTitle = "This person has returned to prison" %}
+
+{% block content %}
+    <div class="govuk-width-container">
+        <main class="govuk-main-wrapper">
+            <div class="govuk-grid-row">
+                <div class="govuk-grid-column-two-thirds">
+                    {{ govukPanel({
+                        titleText: firstName + " " + lastName + " " + "has returned to " + prison.description,
+                        html: "Prison number<br><strong>" + prisonNumber + "</strong>",
```
andrewrlee, line 11: is this prone to html injection? 

The prison number comes from user provided input?

server/routes/temporaryabsences/confirmTemporaryAbsenceAddedToRollController.ts
```diff
@@ -1,9 +1,27 @@
 import { RequestHandler } from 'express'
+import type { PrisonService } from '../../services'
```
andrewrlee, line 2: Might want to add `redirectIfDisabledMiddleware(config.confirmEnabled)` to the GET of this router

server/views/pages/temporaryabsences/confirmTemporaryAbsenceAddedToRoll.njk
andrewrlee: Think it's difficult (maybe impossible to exploit) due to the need to set up flash state before hitting the route.. 

Might be nice to see if there is another way to guarantee it is safe though

server/routes/temporaryabsences/checkTemporaryAbsenceController.test.ts
andrewrlee: Raised DCS-1408 for this

server/views/pages/temporaryabsences/confirmTemporaryAbsenceAddedToRoll.njk
andrewrlee: DCS-1407 for this

server/routes/bookedtoday/index.ts
michalnawrockidj: done

server/routes/temporaryabsences/checkTemporaryAbsenceController.ts
michalnawrockidj: done

server/routes/temporaryabsences/confirmTemporaryAbsenceAddedToRollController.ts
michalnawrockidj: done

michalnawrockidj: We added  if (!firstName || !lastName || !prisonNumber || !prison )  instead

server/routes/temporaryabsences/confirmTemporaryAbsenceAddedToRollController.ts
```diff
@@ -1,9 +1,27 @@
...
   public view(): RequestHandler {
     return async (req, res) => {
-      return res.render('pages/temporaryabsences/confirmTemporaryAbsenceAddedToRoll.njk', {})
+      const { prisonNumber } = req.params
+      const { activeCaseLoadId } = res.locals.user
+      const { firstName, lastName } = req.flash('prisoner')?.[0] as Record<string, string>
+
+      const prison = await this.prisonService.getPrison(activeCaseLoadId)
+
+      if (!firstName || !lastName || !prisonNumber || !prison) {
```
andrewrlee, line 15: meh.. it's not possible to not have a prisonNumber or a prison - the route wouldn't fire if you don't have a prisonNumber, and the getPrison call would throw an exception..

andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

PR #95, opened 2022-01-19
DCS-1285 format changes to temporary absences list

louisenorris:


server/services/temporaryAbsencesService.test.ts
```diff
@@ -22,27 +22,31 @@ describe('Temporary absences service', () => {
...
+      movementDateTime: '2022-01-16T12:30:00',
     },
     {
       firstName: 'Karl',
       lastName: 'Offender',
       dateOfBirth: '1985-01-01',
       prisonNumber: 'G0015GD',
       reasonForAbsence: 'Hospital appointment',
+      movementDateTime: '2022-01-05T10:20:00',
     },
```
louisenorris, line 50: Duplicate lists for temporary absences in both service and controller test files, maybe these should be extracted out to test utils? (unless it is useful to have them present in each test suite to see exactly what's going on). Same list used again for the integration tests in `integration_tests/mockApis/responses/temporaryAbsences.ts` but maybe it is best to keep integration test responses separate from service and controller test data?

andrewrlee (approved):
LGTM

server/services/temporaryAbsencesService.test.ts
andrewrlee: I think this is fine - we likely have the same situation with the arrivals and transfers. We could look at tidying this up but not needed as part of this. 

Regarding int test situation - I've had problems with sharing code across both real and int-test typescript projects so might be difficult (it was a long time ago though)

louisenorris: Ok, cool :)

PR #99, opened 2022-01-25
Dcs 1370 court return check and confirm page

louisenorris:


server/@types/welcome/index.d.ts
```diff
@@ -80,6 +81,9 @@ declare module 'welcome' {
     Prison: {
       description: string
     }
+    PrisonNumber: {
+      prisonNumber: string
+    }
```
louisenorris, line 86: Added above which is effectively duplication of OffenderNumber below as we will be swapping over to using strictly prisonNumber soon and this is being returned from `confirmCourtReturn`. Left offenderNumber as being used in multiple places to be updated at a later date.

andrewrlee (approved):
LGTM

PR #106, opened 2022-02-03
DCS-1326 padding removed so that the title 

GurnankCheema:
'Details from Person Escort Record' is on one line

Discussion with designer confirms the 40/60 column widths are okay


louisenorris (approved):
LGTM

PR #111, opened 2022-02-07
DCS-1407 remove use of html for unsanitised data, escape to explicitly sanitise html where this is not possible

louisenorris:


GurnankCheema (approved):
LGTM

PR #113, opened 2022-02-08
DCS-1348 Adding validation for prisoner name

andrewrlee:


server/routes/__testutils/appSetup.ts
```diff
@@ -24,19 +24,9 @@ export const user = {
 }
 
 export const signedCookiesProvider = jest.fn()
+export const flashProvider = jest.fn()
```
andrewrlee, line 27: I've made one instance to flash provider here so can just import one into the test where needed, don't need to create one per test and inject it in

Want to do the same with roles too.

PR #118, opened 2022-02-09
DCS-1408 refactor of raiseAnalyticsEvent

louisenorris:
We previously directly imported google analytics event publishing where it is used. Refactored this to avoid use of implicit dependencies.

### DIFF
--- a/server/routes/__testutils/appSetup.ts
+++ b/server/routes/__testutils/appSetup.ts
@@ -61,7 +61,6 @@ export function appWithAllRoutes({
   services?: Partial<Services>
   userSupplier?: () => Express.User
   roles?: Role[]
-  raiseAnalyticsEvent?: (category: string, action: string, label: string, hostname: string) => void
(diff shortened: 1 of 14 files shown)


server/routes/__testutils/appSetup.ts
```diff
@@ -61,6 +61,7 @@ export function appWithAllRoutes({
   services?: Partial<Services>
   userSupplier?: () => Express.User
   roles?: Role[]
+  raiseAnalyticsEvent?: (category: string, action: string, label: string, hostname: string) => void
```
andrewrlee, line 64: this shouldn't be needed

louisenorris: eek, sorry fixing now

louisenorris: fixed in latest commit

andrewrlee (approved):
LGTM

PR #130, opened 2022-02-16
DCS-1450 Adding helper method for testing setting of cookies in unit tests

andrewrlee:


server/routes/__testutils/requestTestUtils.ts
```diff
@@ -67,3 +68,14 @@ export const mockResponse = ({ locals = { context: {}, user: {} } }: ResponsePar
   } as unknown as jest.Mocked<Response>)
 
 export const mockNext = (): NextFunction => jest.fn()
+
+export const expectSettingCookie = (res: superAgent.Response, cookieName: string) => {
+  const [cookie] = res.header['set-cookie']
+  const [, name, value] = cookie.match(/^(.*?)=(.*?);/)
```
louisenorris, line 74: not sure if this matters but is there an extra comma here?

andrewrlee: that just means ignore the first value - I'd prefer to set it to a placeholder `_` or something but eslint moans about that.

andrewrlee: `_` doesn't work if you need to do it twice in the same block (in other languages it's a language feature)

louisenorris (approved):
LGTM

PR #151, opened 2022-02-25
DCS-1445: 🔧 align Dockerfile with template project to discourage circle caching image dependencies

connormaglynn:
It looks like CircleCI is always pulling a cache when we run the apt update / upgrade commands, meaning new builds don't pull through the latest versions of packages.

This is an issue since we have security alerts pinging and can't seem to get rid of them.

Not entirely sure on the actual issue here, but the assumption is our Dockerfile is incorrect ( I think maybe having a rm -rf will stop it caching all the commands? ). But I've aligned it more with how the template project does things in hopes that fixes it. Noticed most of our projects build these in different ways, so hard to tell what exactly it is causing the issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,14 +15,17 @@ RUN addgroup --gid 2000 --system appgroup && \
 WORKDIR /app
 
 RUN apt-get update && \
-    apt-get upgrade -y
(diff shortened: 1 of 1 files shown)


andrewrlee (approved):
LGTM

PR #152, opened 2022-02-28
DCS-1470 set state for no match journey and fixed no match redirect o…

louisenorris:
…n confirmation page

### DIFF
--- a/server/routes/bookedtoday/arrivals/checkAnswersController.ts
+++ b/server/routes/bookedtoday/arrivals/checkAnswersController.ts
@@ -61,7 +61,7 @@ export default class CheckAnswersController {
       req.flash('arrivalResponse', {
         firstName: arrival.firstName,
         lastName: arrival.lastName,
-        prisonNumber: arrival.prisonNumber,
+        prisonNumber: arrivalResponse.prisonNumber,
         location: arrivalResponse.location,
       })
       return res.redirect(`/prisoners/${id}/confirmation`)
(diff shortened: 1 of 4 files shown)


server/routes/bookedtoday/arrivals/checkAnswersController.ts
```diff
@@ -61,7 +61,7 @@ export default class CheckAnswersController {
       req.flash('arrivalResponse', {
         firstName: arrival.firstName,
         lastName: arrival.lastName,
-        prisonNumber: arrival.prisonNumber,
+        prisonNumber: arrivalResponse.prisonNumber,
```
louisenorris, line 64: changed to obtain prisonNumber from arrivalResponse as this no match journey has an undefined prisonNumber in arrival object which was causing a redirect to the choose prisoner page in the confirmation controller as prisonNumber was undefined in flash.

andrewrlee: yep that makes sense

server/routes/bookedtoday/arrivals/noRecordFoundController.ts
```diff
@@ -8,6 +10,16 @@ export default class NoRecordFoundController {
...
       const data = await this.expectedArrivalsService.getArrival(id)
+
+      State.newArrival.set(res, {
+        firstName: convertToTitleCase(data.firstName),
+        lastName: convertToTitleCase(data.lastName),
+        dateOfBirth: data.dateOfBirth,
+        sex: data.gender,
+        prisonNumber: data.prisonNumber,
+        pncNumber: data.pncNumber,
+      })
```
andrewrlee, line 21: this should be done in a post handler - same as the single match one should be moved to

andrewrlee (approved):
LGTM

server/routes/bookedtoday/arrivals/noRecordFoundController.ts
andrewrlee: though I guess this will move to the next page so NBD

PR #155, opened 2022-03-03
DCS-1477 Handling rejected promises

andrewrlee:
Not wrapping this in async handlers causes the process to terminate


louisenorris (approved):
LGTM

PR #159, opened 2022-03-07
DCS-1442 jquery ui css COEP fix

louisenorris:


server/middleware/setUpWebSecurity.ts
```diff
@@ -38,6 +38,7 @@ export default function setUpWebSecurity(): Router {
           fontSrc: ["'self'"],
         },
       },
+      crossOriginEmbedderPolicy: true,
```
louisenorris, line 41: This sets the Cross-Origin-Embedder-Policy to require-corp which will only allow loading of resources from same origin, or resources explicitly marked as loadable from another origin using the crossorigin attribute.

andrewrlee (approved):
LGTM

PR #162, opened 2022-03-10
DCS-1487 Fix issue with court transfers

andrewrlee:
Now we no longer merge in matching prisoners into the top level arrival
details we fail to show prisoner info on the court return pages.

This fixes that by retrieving the appropriate prisoner record.

It also fixes a minor issue with the sex controller, where if you
navigated through and back again and through the sex controller it
would ask to provide the arrivals sex (even though it initially was
present)

### DIFF
--- a/server/services/expectedArrivalsService.ts
+++ b/server/services/expectedArrivalsService.ts
@@ -10,6 +10,7 @@ import moment, { type Moment } from 'moment'
 import type { Readable } from 'stream'
 import { groupBy, compareByFullName } from '../utils/utils'
 import type { RestClientBuilder, WelcomeClient, HmppsAuthClient } from '../data'
+import logger from '../../logger'
 
 export enum LocationType {
   COURT = 'COURT',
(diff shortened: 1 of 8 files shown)


server/services/expectedArrivalsService.ts
```diff
@@ -56,6 +57,15 @@ export default class ExpectedArrivalsService {
...
   }
 
+  public async getPrisonerDetailsForArrival(id: string): Promise<PrisonerDetails> {
+    const token = await this.hmppsAuthClient.getSystemClientToken()
+    const arrival = await this.welcomeClientFactory(token).getArrival(id)
+    if (arrival.potentialMatches.length > 1) {
+      logger.warn(`multiple matches for move: ${id}`)
+    }
+    return arrival.potentialMatches[0]
+  }
```
andrewrlee, line 67: this provides the actual prisoner details from nomis associated with the arrival.

This assumes there is only a single potential match

louisenorris (approved):
LGTM

PR #165, opened 2022-03-10
DCS-1390 Tidying up searching for existing record area

andrewrlee:


server/views/pages/bookedtoday/arrivals/searchforexisting/search/changeDateOfBirth.njk
```diff
@@ -0,0 +1,54 @@
+{% extends "../../../../../partials/layout.njk" %}
```
andrewrlee, line 1: split out (duplicated) these as discussed so have different versions between searching and changing details

louisenorris (approved):
LGTM

PR #171, opened 2022-03-15
DCS-1499 corrected search-for-existing-record result pages url paths

louisenorris:
Updated no match page content with Personal details subheading for arrival details. We now get these details from state and make sure we only display name and dob:
Screenshot 2022-03-15 at 14 24 21

Updated single match page content with Personal details subheading for arrival details. 
Screenshot 2022-03-15 at 14 25 42

Also updated url path for match result pages which follow the Search for an existing prisoner record page.


andrewrlee (approved):
LGTM

PR #175, opened 2022-03-16
DCS-1481 Sending from location when confirming arrival

andrewrlee:
Also tidying up terminology around confirming an arrival

Could have handled this in the backend but would have had to make an additional call to BASM and we also need to provide this in the case of unexpected arrivals


GurnankCheema (approved):
LGTM

PR #177, opened 2022-03-17
DCS-1503 clear out new arrival state after no result or multiple results found

andrewrlee:
Fixes an issue where if you:
* Search for a record and produce a match
* Select the match and then go through to check your answers you’ll see the details of the match
* If you navigate back to the search page and alter the query so you produce no match
* Confirm this and progress to the check your answer page you will see the old match

### DIFF
--- a/server/routes/bookedtoday/arrivals/searchforexisting/search/changeDateOfBirthController.test.ts
+++ b/server/routes/bookedtoday/arrivals/searchforexisting/search/changeDateOfBirthController.test.ts
@@ -112,17 +112,13 @@ describe('POST /search-for-existing-record/change-date-of-birth', () => {
       .expect(302)
       .expect('Location', '/prisoners/12345-67890/search-for-existing-record')
       .expect(res => {
-        expect(res.header['set-cookie'][0]).toContain(
-          encodeURIComponent(
-            JSON.stringify({
-              firstName: 'James',
-              lastName: 'Smyth',
-              dateOfBirth: '2003-02-01',
-              prisonNumber: 'A1234AB',
-              pncNumber: '99/98644M',
-            })
-          )
-        )
+        expectSettingCookie(res, 'search-details').toStrictEqual({
+          firstName: 'James',
+          lastName: 'Smyth',
+          dateOfBirth: '2003-02-01',
+          prisonNumber: 'A1234AB',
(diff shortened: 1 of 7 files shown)


server/views/pages/bookedtoday/arrivals/searchforexisting/search/searchForExistingRecord.njk
```diff
@@ -11,8 +11,7 @@
                     valid PNC or prison number.</p>
                 <p>Check their personal details are correct and add a PNC or prison number if you have one. This information will be
                     used to search for an existing prisoner record.</p>
-                <h2 class="govuk-heading-m">Information from the Person Escort Record</h2>
-                <h3 class="govuk-heading-m">Personal details</h3>
```
andrewrlee, line 15: Removing heading as discussed as this data won't be from the PER if the user modifies any data on the page

server/routes/bookedtoday/arrivals/searchforexisting/search/searchForExistingRecordController.ts
```diff
@@ -63,7 +60,12 @@ export default class SearchForExistingRecordController {
 
         return res.redirect(`/prisoners/${id}/search-for-existing-record/record-found`)
       }
-      return res.redirect(`/prisoners/${id}/search-for-existing-record/no-record-found`)
+
+      State.newArrival.clear(res)
```
andrewrlee, line 64: always clear state if there is not a single result

server/routes/bookedtoday/arrivals/searchforexisting/search/changeDateOfBirthController.test.ts
```diff
@@ -112,17 +112,13 @@ describe('POST /search-for-existing-record/change-date-of-birth', () => {
       .expect(302)
       .expect('Location', '/prisoners/12345-67890/search-for-existing-record')
       .expect(res => {
-        expect(res.header['set-cookie'][0]).toContain(
```
andrewrlee, line 115: move to use test util

swilson-mojhub (approved):
Looks good simple text change to the template and can see you are clearing the state if the user has been routed to `no-record-found`

Tests updated

PR #178, opened 2022-03-17
DCS-1504 fix display of sex when single match

andrewrlee:
This previously read the value from the arrival rather than the match

This would only affect the display of the sex in the case it differed or wasn't provided by BASM


petergphillips (dismissed):
LGTM

PR #180, opened 2022-03-18
DCS-1423 Make content more generic

andrewrlee:


Dockerfile
```diff
@@ -14,6 +14,9 @@ RUN addgroup --gid 2000 --system appgroup && \
 
 WORKDIR /app
 
+# Cache breaking
+ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
+
```
andrewrlee, line 19: Migrated fix across from the template project to ensure we always pull in the latest dependencies

server/views/pages/bookedtoday/arrivals/searchforexisting/search/searchForExistingRecord.njk
```diff
@@ -7,10 +7,7 @@
             <div class="govuk-grid-column-two-thirds">
                 <span class="govuk-caption-xl">Confirm a prisoner's arrival</span>
                 <h1 class="govuk-heading-l">{{ pageTitle }}</h1>
-                <p>We could not check if this person has an existing prisoner record because the Person Escort Record does not include a
-                    valid PNC or prison number.</p>
-                <p>Check their personal details are correct and add a PNC or prison number if you have one. This information will be
-                    used to search for an existing prisoner record.</p>
+                <p>Check this person's details are correct and add or change their PNC or prison number if you have one.</p>
```
andrewrlee, line 10: New generic content as this is used both when we have no match and no identifiers and also when we have automatically matched and the user wants to ignore it

GurnankCheema (approved):
LGTM

PR #183, opened 2022-03-22
DCS-1464 cleanup of gender/sex inconsistencies - gender only remaining in arrival object code now as gender rather than sex is used in BASM

louisenorris:


hyderkhanmoj (approved):
LGTM

PR #194, opened 2022-03-28
DCS-1496 add backlink navigation

louisenorris:


server/views/partials/layout.njk
```diff
@@ -66,15 +65,24 @@
...
-
 {% block header %}
   {% include "./header.njk" %}
 {% endblock %}
 
 {% block bodyStart %}
 {% endblock %}
 
+{% block beforeContent %}
+  {% if not noBackLink %}
```
connormaglynn, line 76: minor one you might have already thought about, the double negative is a bit of a code smell since they're hard to read. May be fine if the alternatives are too much work, worth having a think about though 🙂

louisenorris: Hmm, yea agree will sort this asap

integration_tests/pages/page.ts
```diff
@@ -15,12 +15,15 @@ export default abstract class Page {
...
 
-  constructor(private readonly title: string) {
+  constructor(private readonly title: string, private readonly noBackLink: boolean = false) {
     this.checkOnPage()
   }
 
   checkOnPage(): void {
     cy.get('h1').contains(this.title)
+    if (this.noBackLink === false) {
+      cy.get('[data-qa=back-link]').should('exist')
```
connormaglynn, line 25: might be better to just check this in the unit tests instead? Think maybe method where we could check the href in the backlink in the integration tests would be good though? 🙂

server/views/partials/layout.njk
connormaglynn: I think a positive default is one of the exceptions for the smell, not sure how difficult it is assume true within the template if it's not set and have the variable named something like 'displayBackLink' - it's okay to leave if it's a lot of hassle 😄

connormaglynn (approved):
LGTM - just a few minor comments 😁

integration_tests/pages/page.ts
connormaglynn: ^ then again might be overkill to check the href in multiple places since we're using the inbuilt history function in js. Suppose it'd make the tests more valuable if we change that implementation later on, but we could cross that bridge when we come to it 😄

server/views/partials/layout.njk
```diff
@@ -73,7 +73,7 @@
 {% endblock %}
 
 {% block beforeContent %}
-  {% if not noBackLink %}
+  {% if backLink !== false %}
```
louisenorris, line 76: Hopefully an improvement on the double negative - need to check it is not false rather than true as not setting the variable as true globally so I can override as false in confirmation pages.

integration_tests/pages/page.ts
andrewrlee: Hey @connormaglynn , We talked about this on Friday and we don't have a way of verifying for every page in unit tests and can't test javascript so thought it might be better to just handle in integration tests

integration_tests/integration/bookedtoday/arrivals/autoMatchingRecords/isSingleMatch.spec.ts
```diff
@@ -113,4 +113,29 @@ context('Is Single Match', () => {
...
+
+    const imprisonmentStatusPage = Page.verifyOnPage(ImprisonmentStatusPage)
+    imprisonmentStatusPage.backNavigationLink().should('exist')
+    imprisonmentStatusPage.imprisonmentStatusRadioButton('on-remand').click()
+    imprisonmentStatusPage.continue().click()
+
+    const checkAnswersPage = Page.verifyOnPage(CheckAnswersPage)
+    checkAnswersPage.backNavigationLink().should('exist')
+    cy.task('stubCreateOffenderRecordAndBooking', { arrivalId: expectedArrival.id })
+    checkAnswersPage.addToRoll().click()
```
andrewrlee, line 136: could maybe test back link here and go back a couple pages then click through to check back not being present

integration_tests/pages/authManageDetails.ts
```diff
@@ -2,6 +2,6 @@ import Page from './page'
 
 export default class AuthManageDetailsPage extends Page {
   constructor() {
-    super('Your account details')
+    super('Your account details', false)
```
andrewrlee, line 5: think this is fine but just to be aware having [boolean params](https://martinfowler.com/bliki/FlagArgument.html) is often seen as a code smell for a few reasons, one being that it's not obvious when reading the code what they mean.  

A way to help in this situation might be passing an object as the boolean then has context of having a  name
e.g:
```
 super('Your account details', { hasBackLink: false })
```

integration_tests/pages/page.ts
```diff
@@ -15,14 +15,19 @@ export default abstract class Page {
...
   }
 
-  constructor(private readonly title: string) {
+  constructor(private readonly title: string, private readonly backLink: boolean = true) {
     this.checkOnPage()
   }
 
   checkOnPage(): void {
     cy.get('h1').contains(this.title)
+    if (this.backLink !== false) {
```
andrewrlee, line 24: maybe should also test that the link doesn't exist when it shouldn't?

andrewrlee: and should be able to use `this.backNavigationLink`?

andrewrlee (approved):
lgtm

integration_tests/pages/page.ts
```diff
@@ -15,14 +16,21 @@ export default abstract class Page {
...
+  constructor(private readonly title: string, private readonly backNavigationLink: BackLink = { hasBackLink: true }) {
     this.checkOnPage()
   }
 
   checkOnPage(): void {
     cy.get('h1').contains(this.title)
+    if (this.backNavigationLink.hasBackLink !== false) {
+      cy.get('[data-qa=back-link]').should('exist')
+    } else {
+      cy.get('[data-qa=back-link]').should('not.exist')
```
andrewrlee, line 28: meh.. could reuse `this.backNavigation()` in these couple of places

server/views/pages/temporaryabsences/confirmTemporaryAbsenceAddedToRoll.njk
```diff
@@ -1,4 +1,5 @@
 {% extends "../../partials/layout.njk" %}
+{% set backNavigationLink = { hasBackLink: false } %}
```
andrewrlee, line 2: very meh.. but worth pointing out, here you don't have the same thing with primitives - the variable name can provide the context you need, so you could get away with:
```
{% set hasBackLink = false %}
```

as where you use it can refer to that name:

```
{% if hasBackLink !== false %}
```

integration_tests/pages/page.ts
louisenorris: thanks - sorted in latest commit

server/views/pages/temporaryabsences/confirmTemporaryAbsenceAddedToRoll.njk
louisenorris: thanks - sorted in latest commit

andrewrlee (approved):
LGTM

PR #202, opened 2022-04-01
DCS-1466 add link on prisoner search page

GurnankCheema:
Link takes user /manually-confirm-arrival/search-for-existing-record
Only visible if user has reception role


andrewrlee (approved):
LGTM

PR #203, opened 2022-04-01
DCS-1539 change link target

GurnankCheema:
link to go to new search rather than create record.
see /v13/confirm-arrival/new-arrival/record-found of the prototype


andrewrlee (approved):
LGTM

PR #208, opened 2022-04-05
DCS-1553 Prevent passing of prison number on new record

andrewrlee:
This would only happen if we received an invalid prison number from BASM. We were previously passing it through even though it must have been invalid else we would have received a match


yasinmustafa (approved):
LGTM

PR #213, opened 2022-04-08
DCS-1516 Adding integration tests

andrewrlee:


integration_tests/pages/bookedtoday/arrivals/confirmArrival/confirmAddedToRoll.ts
```diff
@@ -9,7 +13,22 @@ export default class ConfirmAddedToRollPage extends Page {
...
+    prison: string
+    locationName: string
+    prisonNumber: string
+    name: string
+  }) => {
+    cy.get(paragraph).contains(prison)
+    cy.get(location).contains(locationName)
+    cy.get(banner).contains(prisonNumber)
+    cy.get(banner).contains(name)
+  }
```
andrewrlee, line 31: must be a nicer way to do this rather than getting the elements multiple times - cypress lint doesn't like assigning to variables so think need to use promise callback but then it gives you a jquery element rather than cypress element  - nevertheless this works and is equivalent to what was being called before

server/services/expectedArrivalsService.ts
```diff
@@ -79,39 +71,68 @@ export default class ExpectedArrivalsService {
     const token = await this.hmppsAuthClient.getSystemClientToken(username)
     const welcomeClient = this.welcomeClientFactory(token)
 
+    return arrival.expected
+      ? this.confirmExpectedArrival(welcomeClient, prisonId, id, arrival)
+      : this.confirmUnexpectedArrival(welcomeClient, prisonId, arrival)
+  }
```
andrewrlee, line 77: little bit of refactoring here as we were trying to look up an unexpected arrival when by definition we don't have an arrival record for them.

louisenorris (approved):
LGTM

PR #214, opened 2022-04-08
DCS-1516 Extracting out factories to reuse test data

andrewrlee:


integration_tests/integration/bookedtoday/choosePrisoner.spec.ts
```diff
@@ -22,7 +22,6 @@ context('Choose Prisoner', () => {
         expectedArrivals.custodySuite.current,
         expectedArrivals.custodySuite.notCurrent,
         expectedArrivals.custodySuite.notMatched,
-        expectedArrivals.other,
```
andrewrlee, line 25: not used

server/data/__testutils/testObjects.ts
```diff
@@ -0,0 +1,147 @@
...
+  lastName = 'Smith',
+  dateOfBirth = '1973-01-08',
+  prisonNumber = 'A1234AB',
+  pncNumber = '01/98644M',
+  date = '2021-10-13',
+  fromLocation = 'Reading Court',
+  fromLocationType = 'COURT',
+  isCurrentPrisoner = true,
+  potentialMatches = [createPotentialMatch()],
+} = {}): Arrival => ({
```
andrewrlee, line 26: these test data factories all allow you to create a default test object but allow you to override specific fields for required tests. 
Should only need to override specific fields if the test has logic that requires overriding

server/data/welcomeClient.test.ts
```diff
@@ -51,52 +52,44 @@ describe('welcomeClient', () => {
...
     const prisonNumber = 'A1234AB'
-    const expectedArrival: Transfer = {
-      firstName: 'Sam',
-      lastName: 'Smith',
-      dateOfBirth: '1971-02-01',
-      prisonNumber: 'A1234AA',
-      pncNumber: '01/1234X',
-      date: '2020-02-23',
-      fromLocation: 'Kingston-upon-Hull Crown Court',
-    }
```
andrewrlee, line 91: for instance - these tests don't require the transfer to have any specific values - it just requires a valid one.

We can completely change all the values with out breaking the test.

We want to limit  the coupling of test data in tests  - if we change the default values in the factories then it should break few if any tests

hyderkhanmoj (approved):
LGTM

PR #217, opened 2022-04-13
DCS-1541 handle display of prisoner search details

GurnankCheema:
If either of prisoner name, dob, prison number, pnc are not provided in search the text 'Not entered' will be displayed.
If the first name is not provided but last name is, then first name will not be displayed. 
Previously the above missing criteria would result in the ui displaying 'undefined'


andrewrlee (approved):
LGTM

PR #220, opened 2022-04-13
DCS-1463 moving away from deprecated endpoints

andrewrlee:


server/data/restClient.ts
```diff
@@ -45,7 +45,7 @@ export default class RestClient {
   }
 
   async get({ path = null, query = {}, headers = {}, responseType = '', raw = false }: GetRequest): Promise<unknown> {
-    logger.info(`Get using user credentials: calling ${this.name}: ${path} ${query}`)
+    logger.info(`Get using user credentials: calling ${this.name}: ${path}`)
```
andrewrlee, line 48: this logging only every printed `{ Object object }` so removing

GurnankCheema (approved):
LGTM

PR #229, opened 2022-04-19
DCS-1507: ✨ Add the split information screen for court returns

connormaglynn:


server/views/pages/bookedtoday/arrivals/autoMatchingRecords/singleMatchingRecordFound.njk
```diff
@@ -1,39 +1,24 @@
...
-                {{ prisonerSummaryDetail(arrival, "per-record" ) }}
-            </div>
-        </div>
-        <div class=" border-left-grey govuk-!-padding-left-9 column-width-60">
-            <h2 class="govuk-heading-m">Details from existing prisoner record</h2> 
-            <div class="prisoner-record-detail flex">
-                {{ prisonerSummaryDetail(data, "existing-record") }}
-                <img src="{{ '/prisoners/' + data.potentialMatches[0].prisonNumber + '/image' if data.potentialMatches[0].prisonNumber else '/assets/images/placeholder-image.png' }}" 
-                        class="prisoner-image-transfer govuk-!-margin-top-2" 
-                        alt="{{ data.potentialMatches[0].lastName }}, {{ data.potentialMatches[0].firstName }}"
```
connormaglynn, line 26: think this might have been a bug, doesn't look like we pass a potentialMatch to the view?

server/views/components/prisoner-split-view.njk
```diff
@@ -0,0 +1,24 @@
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+{% from "./prisoner-summary-detail.njk" import prisonerSummaryDetail %}
+
+{% macro prisonerSplitView(prisonerEscortRecord, prisonRecord, personDetailsHeaderText='Details from Person Escort Record') %}
```
connormaglynn, line 4: Seemed like a good candidate for it's own component since the majority of the code is the same. The only one I didn't move over was the unexpected single match since it used prisonerSearchDetail() instead of prisonerSummaryDetail()

server/routes/bookedtoday/arrivals/courtreturns/checkCourtReturnController.test.ts
```diff
@@ -52,6 +53,20 @@ describe('checkCourtReturnController', () => {
         .expect(res => {
           const $ = cheerio.load(res.text)
           expect($('h1').text()).toContain('This person is returning from court')
+          expect($('.data-qa-per-record-prisoner-name').text()).toContain(`${arrival.firstName} ${arrival.lastName}`)
```
connormaglynn, line 56: Didn't put this in the int tests since it's tested here. Thought about moving it into the commonTests.ts file, but got a bit bogged down with the typing 😆

andrewrlee: This stuff should really go in an int test - don't really want these tests super coupled tightly to the view which is why we generally only check the title.

I've been stripping out this from other controller tests

server/views/pages/bookedtoday/arrivals/autoMatchingRecords/singleMatchingRecordFound.njk
andrewrlee: good spot

server/views/pages/bookedtoday/arrivals/searchforexisting/singleExistingRecordFound.njk
```diff
@@ -40,11 +24,12 @@
...
-                    <a class="govuk-body govuk-link govuk-link--no-visited-state" href="/confirm-arrival/choose-prisoner" data-qa="choose-prisoner">
+                    <a class="govuk-body govuk-link govuk-link--no-visited-state"
+                       href="/confirm-arrival/choose-prisoner" data-qa="choose-prisoner">
                         Return to list of prisoners arriving today
                     </a>
-                </p> 
-            </div>   
+                </p>
+            </div>
         </main>
```
andrewrlee, line 33: worth checking the html validator when modifying pages - this won't validate due to multiple main elements

server/views/pages/bookedtoday/arrivals/courtreturns/checkCourtReturn.njk
```diff
@@ -83,14 +31,14 @@
...
+                                href="/confirm-arrival/choose-prisoner"
+                                data-qa="return-to-arrivals-list">
                             Return to list of prisoners arriving today
                         </a>
-                    </p>   
+                    </p>
                 </div>
-            </div> 
+            </div>
         </main>
```
andrewrlee, line 42: same here - re: html validation

server/views/components/prisoner-split-view.njk
andrewrlee: Would be tempted to not default the heading and make it a mandatory argument - when its only 2 occurrences vs 1 there's not really a clear default.

integration_tests/components/prisonerSplitView.ts
```diff
@@ -0,0 +1,19 @@
+import { PageElement } from '../pages/page'
+
+export default class PrisonerSplitViewComponent {
```
connormaglynn, line 3: Added the concept of components into the integration tests. The reasoning is I think we could build up pages through the composition of components rather than defining each attribute to lessen some duplication and add some standardisation. Thoughts?

andrewrlee (approved):
LGTM!

PR #231, opened 2022-04-20
DCS-1574 validation for missing last name

GurnankCheema:
Validation prevents a search when either first name or dob are present but last name is missing.
This also means, the last name will not be displayed as 'undefined' in the search results page

### DIFF
--- a/server/routes/bookedtoday/unexpectedArrivals/validation/searchForExistingRecordsValidation.test.ts
+++ b/server/routes/bookedtoday/unexpectedArrivals/validation/searchForExistingRecordsValidation.test.ts
@@ -17,6 +17,16 @@ describe('MatchedRecordSelectionValidation', () => {
       },
     ]))
 
+  it('Should return error if only DOB and prison number', () =>
+    expect(
+      SearchForExistingRecordsValidation({ day: '02', month: '01', year: '2000', prisonNumber: 'A1234BC' })
+    ).toEqual([
+      {
+        href: '#last-name',
+        text: "You must search using either the prisoner's last name and date of birth, prison number or PNC Number",
+      },
+    ]))
+
   it('Should return error if only first name entered', () =>
     expect(SearchForExistingRecordsValidation({ firstName: 'James' })).toEqual([
       { href: '#last-name', text: 'Enter a last name and a date of birth' },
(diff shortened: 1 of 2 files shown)


server/routes/bookedtoday/unexpectedArrivals/validation/searchForExistingRecordsValidation.test.ts
```diff
@@ -20,4 +21,19 @@ describe('MatchedRecordSelectionValidation', () => {
     expect(SearchForExistingRecordsValidation({ firstName: 'James' })).toEqual([
       { href: '#last-name', text: 'Enter a last name and a date of birth' },
     ]))
+
```
GurnankCheema, line 24: tests ordered as closely as possible to the rules in the 'rules' spreadsheet

server/routes/bookedtoday/unexpectedArrivals/validation/searchForExistingRecordsValidation.ts
```diff
@@ -9,11 +9,15 @@ const SearchForExistingRecordsValidator: Validator = ({
...
+  if (firstName && !lastName && !day && !month && !year) {
     return [{ text: 'Enter a last name and a date of birth', href: '#last-name' }]
   }
 
-  if (!lastName && !prisonNumber && !pncNumber) {
+  if (firstName && !lastName) {
+    return [{ text: 'Enter a last name', href: '#last-name' }]
+  }
+
+  if ((!firstName && !lastName && day && month && year) || (!lastName && !prisonNumber && !pncNumber)) {
```
andrewrlee, line 20: can this can just be `(!lastName && !prisonNumber && !pncNumber)` ? 

Wouldn't the other case be covered anyway?

GurnankCheema: yes it can be.

GurnankCheema: No because it would allow a search using just dob plus either prison or pnc number

andrewrlee: cool, might be nice to have this as a parameterised test so we can see that every combo has been handled. 

there's no tests for valid cases for instance

andrewrlee (approved):
LGTM

PR #236, opened 2022-04-27
DCS-1581 handle client errors for transfers

GurnankCheema:
400-500 client errors when calling welcome-api will cause null to be returned to the controller which which will then redirect to /feature-not-available


andrewrlee (approved):
LGTM

PR #247, opened 2022-05-18
DCS-1618 Updating dependencies, bumping version of orb and node

andrewrlee:


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@3.12
+  hmpps: ministryofjustice/hmpps@5
```
mikehalmamoj, line 4: I thought about doing this too - decided it's a bit non-determinstic 🤷 Though it should be fine!

mikehalmamoj (approved):

PR #248, opened 2022-05-19
DCS-1598 stub and show recent arrivals sorted by datetime

louisenorris:


server/views/partials/expectedArrivalCardMacro.njk
```diff
@@ -39,6 +39,43 @@
     </dl>
 {% endmacro %}
 
+{% macro recentArrivalCardListContent(arrival, loopIndex) %}
```
andrewrlee, line 42: maybe these should go in `recentArrivalCardMacros`?

server/views/pages/recentArrivals/recentArrivals.njk
```diff
@@ -1,16 +1,29 @@
...
     <div class="govuk-!-margin-bottom-8">
         <div class="govuk-width-container govuk-body">
             <h1>{{pageTitle}}</h1>
-                <h2 class="govuk-heading-m govuk-!-margin-top-9">Wednesday 27 April</h2>
-
-                <h2 class="govuk-heading-m govuk-!-margin-top-9">Tuesday 26 April</h2>
-
-                <h2 class="govuk-heading-m govuk-!-margin-top-9">Monday 25 April</h2>
+            <h2 class="govuk-heading-m govuk-!-margin-top-9" id='today'>{{dates.today}}</h2>
+                {% for arrival in recentArrivals.get(dates.today) %}
```
andrewrlee, line 10: might be tempted to push this down into the service layer and just iterate over the `Map<Date, List<Move>>`?

The date calculation in the controller could be pushed down too - that will probably help if we need to change this.

server/utils/utils.ts
```diff
@@ -36,6 +40,10 @@ export const compareByFullName = <T extends Person>(a: T, b: T): number => {
   return result !== 0 ? result : a.firstName.localeCompare(b.firstName)
 }
 
+export const compareByDateAndTime = <T extends DateAndTime>(a: T, b: T): number => {
+  return moment(a.movementDateTime).unix() - moment(b.movementDateTime).unix()
+}
```
andrewrlee, line 45: this is a little bit of the domain creeping into otherwise domain agnostic code.

could possibly handle this by saying you have something of type `T` 
and a mechanism (function) of getting a date from it: `T => Date`

server/services/expectedArrivalsService.ts
```diff
@@ -28,13 +36,27 @@ export default class ExpectedArrivalsService {
...
     const transfers = await welcomeClient.getTransfers(agencyId)
     return transfers.map(transfer => ({ ...transfer, fromLocationType: LocationType.PRISON })).sort(compareByFullName)
   }
 
+  public async getRecentArrivalsGroupedByDate(agencyId: string): Promise<Map<string, RecentArrival[]>> {
+    const recentArrivals = await this.getRecentArrivals(agencyId)
+    return groupBy(recentArrivals, (recentArrival: RecentArrival) =>
+      moment(recentArrival.movementDateTime).format('dddd D MMMM')
+    )
+  }
```
andrewrlee, line 58: sweet -  if we were to push stuff down here, we'd have to do something funky creating a range of dates

server/views/partials/expectedArrivalCardMacro.njk
louisenorris: sorted in latest PR

server/views/pages/recentArrivals/recentArrivals.njk
louisenorris: sorted in latest PR

server/utils/utils.ts
louisenorris: sorted in latest PR

GurnankCheema (approved):
LGTM

PR #249, opened 2022-05-20
DCS-1616: ✨ Add user to Request in App insights

connormaglynn:


server/utils/azureAppInsights.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  activeCaseLoadDescription: 'Lincoln',
+}
+
+const createEnvelope = (properties: Record<string, string | boolean>, baseType = 'RequestData') =>
+  ({
+    data: {
+      baseType,
+      baseData: { properties },
+    },
+  } as unknown as EnvelopeTelemetry)
```
connormaglynn, line 18: the as unknown as <TYPE> feels a bit hacky, anyone know a better way to quickly define a type without having to specify the entire structure?

server/utils/azureAppInsights.ts
```diff
@@ -29,7 +31,29 @@ export function buildAppInsightsClient(name = defaultName()): TelemetryClient {
...
+    const { username, activeCaseLoadId, activeCaseLoad, isReceptionUser } =
+      contextObjects?.['http.ServerRequest']?.res?.locals?.user || {}
+    const { properties } = envelope.data.baseData
+    // eslint-disable-next-line no-param-reassign
+    envelope.data.baseData.properties = {
+      username,
+      activeCaseLoadId,
+      activeCaseLoadDescription: activeCaseLoad?.description,
+      isReceptionUser,
+      ...properties,
```
connormaglynn, line 55: Not sure if this is a bit too much info, thinking in terms of PII. I had a quick google and can't seem to say whether it is or isn't, seems fairly subjective in terms of what data could identify someone. They seem like useful data points though whilst we're rolling out

server/utils/azureAppInsights.test.ts
andrewrlee: it usually works without `as unknown` if it contains a subset of the object. This suggests maybe the typing is a bit off, things in the wrong place?

server/utils/azureAppInsights.ts
andrewrlee: think that's fine but can we lose the prison description - quite a lot of additional data storage if you thinks it for every single request and its trivial to do the look up externally.

andrewrlee: good to add the is reception user bit though - it will allow us to see if its regularly useful info for non-reception staff which is just an assumption at the moment

server/utils/azureAppInsights.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  username: 'test-user',
+}
+
+const createEnvelope = (properties: Record<string, string | boolean>, baseType = 'RequestData') =>
+  ({
+    data: {
+      baseType,
+      baseData: { properties },
+    } as DataTelemetry,
+  } as EnvelopeTelemetry)
```
andrewrlee, line 16: oh nice! that's interesting that it can't cope with the partial object at the lower level

server/utils/azureAppInsights.ts
```diff
@@ -1,7 +1,13 @@
 import { config } from 'dotenv'
-import { setup, defaultClient, TelemetryClient, DistributedTracingModes } from 'applicationinsights'
+import { Contracts, defaultClient, DistributedTracingModes, setup, TelemetryClient } from 'applicationinsights'
+import { EnvelopeTelemetry } from 'applicationinsights/out/Declarations/Contracts'
 import applicationVersion from '../applicationVersion'
 
+export type ContextObject = {
+  /* eslint-disable  @typescript-eslint/no-explicit-any */
+  [name: string]: any
```
andrewrlee, line 8: This is fine but wonder if you could ditch the warning by using unknown.. not sure but maybe its type safe to write to an unknown

andrewrlee (approved):
LGTM

server/utils/azureAppInsights.test.ts
connormaglynn: Yeah thought it would have assumed it as well, so when intellij was telling me to define all the other fields thought that that was the standard process. Think your comment that if they are subset it should work - so thought maybe defining the subset would help it along the way 😁

PR #254, opened 2022-06-01
DCS-1600 added recent arrival search controller and views

louisenorris:
Might have to change the way this is implemented with flash (see note in ticket [DCS-1600](https://dsdmoj.atlassian.net/browse/DCS-1600?focusedCommentId=218436)). Currently storing search query in flash so when a user refreshes the search page we loose the search query and as a result all recent arrivals are shown in the search results list (see screenshot below). For now I’ve added a redirect to take the user back to `/recent-arrivals` when this happens. Spoke to Charan and this isn't ideal. 

Screenshot 2022-06-01 at 18 02 34

Maybe we use a cookie instead or have the query in the URL like in DPS global search. There when you click on the survey monkey link the prison number is used in the page URL, replacing name when a user has searched by name - so maybe this is also a solution to not including PII in URLs. 

I've tried implementing a cookie to fix the above issue locally and it seems to work. Can push this up for consideration once this PR is merged.


GurnankCheema (dismissed):
LGTM

GurnankCheema (approved):
LGTM

server/routes/recentArrivals/index.ts
```diff
@@ -2,11 +2,20 @@ import { Router } from 'express'
...
 
 export default function routes(services: Services): Router {
   const recentArrivals = new RecentArrivalsController(services.expectedArrivalsService)
+  const recentArrivalsSearch = new RecentArrivalsSearchController(services.expectedArrivalsService)
 
-  return Routes.forAnyRole().get('/recent-arrivals', recentArrivals.view()).forRole(Role.PRISON_RECEPTION).build()
+  return Routes.forAnyRole()
+    .get('/recent-arrivals', recentArrivals.view())
+    .post('/recent-arrivals', recentArrivals.search())
+    .forRole(Role.PRISON_RECEPTION)
```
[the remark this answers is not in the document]
andrewrlee: Presumably we would want to allow anybody carry out searches? Or we'd end up having to have to only present the search button if the user had this role

server/routes/recentArrivals/recentArrivalsController.test.ts
```diff
@@ -67,3 +67,16 @@ describe('GET /recent-arrivals', () => {
...
+describe('POST /recent-arrivals', () => {
+  const flash = flashProvider.mockReturnValue([])
+  it('should store search query in flash and redirect to /recent-arrivals/search', () => {
+    return request(app)
+      .post('/recent-arrivals')
+      .expect(302)
+      .expect('Location', '/recent-arrivals/search')
+      .expect(res => {
+        expect(flash).toBeCalledTimes(1)
+      })
```
andrewrlee, line 80: sounds like this test should demo sending the search query and should show what flash is being called with

integration_tests/integration/recentArrivals/recentArrivals.spec.ts
```diff
@@ -64,4 +69,75 @@ context('A user can view all recent arrivals', () => {
...
+
+    recentArrivalsPage.searchInput().type('Mark')
+    cy.task('stubRecentArrivals', {
+      caseLoadId: 'MDI',
+      recentArrivals: recentArrivalsSearchResponse,
+    })
+    recentArrivalsPage.searchSubmit().click()
+    const recentArrivalsSearchPage = Page.verifyOnPage(RecentArrivalsSearchPage)
+
+    recentArrivalsSearchPage.noResultsFound().should('contain.text', `No results found for 'Mark'.`)
```
andrewrlee, line 141: should we be checking we don't see this message in the other situation?

server/routes/recentArrivals/recentArrivalsSearchController.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      .get('/recent-arrivals/search')
+      .expect('Content-Type', 'text/html; charset=utf-8')
+      .expect(res => {
+        const $ = cheerio.load(res.text)
+        expect($('h1').text()).toContain('Prisoners who have arrived in the last 3 days')
+        expect($('.title-1').text()).toContain('Smith, Sam')
+      })
+  })
+
+  it('should display alternative text if no search results to display', () => {
```
andrewrlee, line 61: probably best to avoid testing the same thing at multiple levels, here and unit tests

integration_tests/integration/recentArrivals/recentArrivals.spec.ts
```diff
@@ -64,4 +69,75 @@ context('A user can view all recent arrivals', () => {
         expect(altText).equal('Headshot of Doe, John')
       })
   })
+
+  it('A user can successfully search for a recent arrival', () => {
+    const fromDate = moment().subtract(2, 'days').format('YYYY-MM-DD')
+    const toDate = moment().format('YYYY-MM-DD')
+    const recentArrivalsSearchResponse1 = recentArrivalsResponse.arrivals({
+      content: [
```
andrewrlee, line 77: could push wrapper into the test data factory

PR #259, opened 2022-06-08
DCS-1651 Extending timeouts to calls for pages displaying lists and disabling retries

andrewrlee:
Disabling retries and extending timeouts so we don’t batter BASM as much when they’re having perf issues. 

Also means patient users are more likely to see results quicker!

### DIFF
--- a/server/data/welcomeClient.ts
+++ b/server/data/welcomeClient.ts
@@ -31,6 +31,8 @@ export default class WelcomeClient {
     return this.restClient.get({
       path: `/prisons/${agencyId}/arrivals`,
       query: { date: date.format('YYYY-MM-DD') },
+      timeout: 15000,
+      retryCount: 0,
     }) as Promise<Arrival[]>
   }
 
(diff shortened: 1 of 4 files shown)


server/data/welcomeClient.ts
```diff
@@ -55,6 +57,8 @@ export default class WelcomeClient {
         toDate: toDate.format('YYYY-MM-DD'),
         query: searchQuery,
       },
+      timeout: 15000,
+      retryCount: 0,
     }) as Promise<PaginatedResponse<RecentArrival>>
```
andrewrlee, line 62: default timeout is 5 seconds and 3 retries. 

When BASM "struggles" the highest duration we've seen is 8 seconds. 
This would trigger a retry so the user is unlikely to see anything for at least 15 seconds. 

For this situation, the new config values would mean that the user sees the list in 8 seconds. 
Disabling retries as a user is reasonably likely to have manually retried after 15 seconds.

louisenorris (approved):
LGTM

PR #261, opened 2022-06-14
DCS-1663 disabling double click on buttons

andrewrlee:
In production had issues with a user double clicking when confirming a recall


connormaglynn (approved):
lgtm

PR #267, opened 2022-06-30
DCS-1703 future dob validation

louisenorris:
Past DOB validation added to make sure that a user doesn't accidentally input a date of birth which is in the future in error: 

Screenshot 2022-07-01 at 17 09 16

### DIFF
--- a/server/utils/utils.test.ts
+++ b/server/utils/utils.test.ts
@@ -1,3 +1,4 @@
+import moment from 'moment'
 import {
   convertToTitleCase,
   trimObjectValues,
@@ -9,6 +10,7 @@ import {
   assertHasOptionalStringValues,
   zip,
   isValidDate,
+  isPastDate,
 } from './utils'
 
 describe('Convert to title case', () => {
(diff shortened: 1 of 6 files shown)


server/utils/utils.test.ts
```diff
@@ -396,3 +398,23 @@ describe('isValidDate', () => {
...
+
+  it('has past date', () => {
+    const yesterday = moment().subtract(1, 'days')
+    const day = yesterday.format('DD')
+    const month = yesterday.format('MM')
+    const year = yesterday.format('YYYY')
+    const result = isPastDate(day, month, year)
+    expect(result).toEqual(false)
+  })
+})
```
louisenorris, line 420: Unsure whether we should be testing for today's date as that would surely be input in error by a user too as a date of birth. However, maybe it's not strictly covered by this validation and we should discuss as a team if we need validation to check the minimum age of a person?

connormaglynn: Agree with this, not sure if NOMIS would allow it in - but we probably should be gatekeeping any policy requirements at our front door to stop any issues later down the line. Having a quick google looks like you can be criminally charged at 10, but it also looks like YOI's are geared up for 15 - 18 year olds. We'd need to be careful of using arbitrary rules. So one worth discussing with the team as you said 😃

connormaglynn (approved):
lgtm

PR #279, opened 2022-07-28
DCS-1725 Moving bodyscan stuff into separate folder structure

andrewrlee:
We want to keep all the body scan isolated so it can be migrated to a new service once the domain it belongs to has been identified


server/services/index.ts
```diff
@@ -22,6 +22,7 @@ export const services = () => {
   const prisonService = new PrisonService(hmppsAuthClient, welcomeClientBuilder)
 
   return {
+    hmppsAuthClient,
```
andrewrlee, line 25: sharing `hmppsAuthClient` across both sets of services - we only have one token store/redis client

michalnawrockidj (approved):
LGTM

davidthomas22 (commented):
LGTM

PR #298, opened 2022-08-18
DCS-1691: ♻️ Move to using DPS backlink endpoint for Case Notes

connormaglynn:
PR to update the **Add a case note** link to use the DPS backlink URL so users aren't dead-ended after adding a case note.

There are essentially only two link updates here
- Update Temporary Absence Confirmation so the return URL is **/**
- Update every other Confirmation Page so the return URL is **/confirm-arrival/choose-prisoner**

Apologies for the mass indentations, seems like our standard is four indents hadn't been applied in some instances 👀 I'd recommend hiding whitespace changes in Github to see the actual changes in the .njk files😄


louisenorris (approved):
LGTM

PR #299, opened 2022-08-18
DCS 1700 Add feedback banner component

davidthomas22:


server/views/components/feedback-banner.njk
```diff
@@ -0,0 +1,11 @@
...
+<div class="govuk-!-display-none-print {{ params.feedbackBannerAdditionalClasses }}">
+  <div class="govuk-width-container">
+    <div class="feedback-banner" data-test="feedback-banner" data-qa="feedback-link">
+      <a href="/feedback" class="govuk-link govuk-link--inverse" rel="noopener noreferrer">
+        Suggest an improvement or report a problem with this service
+      </a>
+    </div>
+  </div>
+</div>
+{% endmacro %}
```
louisenorris, line 11: Think we should add an empty lint at the end of this file to avoid the Nunjucks format error on line 11

server/views/partials/footer.njk
```diff
@@ -0,0 +1,29 @@
...
+{% include "../components/feedback-banner.njk" %}
+
+{{ 
+  govukFooter({
+    meta: {
+      items: [
+        {
+          href: '/feedback-and-support',
+          text: "Get help",
+          attributes: {
```
louisenorris, line 21: This "Get help" link was added to BVL but shouldn't be in WPIP... so this govukFooter should be removed I think

integration_tests/pages/bodyscan/bodyScanConfirmation.ts
```diff
@@ -2,7 +2,7 @@ import Page, { PageElement } from '../page'
 
 export default class BodyScanConfirmation extends Page {
   constructor() {
-    super('Body scan recorded', { hasBackLink: false })
+    super('Body scan recorded', { hasBackLink: false }, { hasFeedbackBanner: true })
   }
```
louisenorris, line 6: If we're importing `Page` is `hasFeedbackBanner: true` not the default? Not sure this is needed here but could be wrong.

integration_tests/pages/homePage.ts
```diff
@@ -2,7 +2,7 @@ import Page, { PageElement } from './page'
 
 export default class HomePage extends Page {
   constructor() {
-    super('Welcome people into prison', { hasBackLink: false })
+    super('Welcome people into prison', { hasBackLink: false }, { hasFeedbackBanner: true })
```
louisenorris, line 5: If we're importing `Page` is `hasFeedbackBanner: true` not the default? Not sure this is needed here but could be wrong.

integration_tests/pages/bodyscan/bodyScanConfirmation.ts
```diff
@@ -2,7 +2,7 @@ import Page, { PageElement } from '../page'
 
 export default class BodyScanConfirmation extends Page {
   constructor() {
-    super('Body scan recorded', { hasBackLink: false })
+    super('Body scan recorded', { hasBackLink: false }, { hasFeedbackBanner: true })
```
louisenorris, line 5: If we're importing `Page` is `hasFeedbackBanner: true` not the default? Not sure this is needed here but could be wrong.

server/views/partials/footer.njk
```diff
@@ -12,18 +12,5 @@
...
-          text: "Get help",
-          attributes: {
-                'data-qa': 'get-help'
-            }
-        }
-      ]
-    },
-    classes: "govuk-!-display-none-print qa-court-service-footer"
-  }) 
+  govukFooter()
```
louisenorris, line 15: Just checking, do we need this now? Does the gov footer not show at all if we don't add this? Just checking as we could see the footer previously without it but other changes may mean it's now necessary.

louisenorris (approved):
LGTM

PR #313, opened 2022-09-05
DCS-1668 Updating redirect logic

andrewrlee:
Removing special handling of active prisoners from PCS

Also tidying up court return logic

### DIFF
--- a/integration_tests/integration/bookedtoday/choosePrisoner.cy.ts
+++ b/integration_tests/integration/bookedtoday/choosePrisoner.cy.ts
@@ -1,5 +1,4 @@
 import ChoosePrisonerPage from '../../pages/bookedtoday/choosePrisoner'
-import FeatureNotAvailablePage from '../../pages/featureNotAvailable'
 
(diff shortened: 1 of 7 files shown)


integration_tests/integration/bookedtoday/choosePrisoner.cy.ts
```diff
@@ -166,18 +170,6 @@ context('Choose Prisoner', () => {
     Page.verifyOnPage(SingleMatchingRecordFoundPage)
   })
 
-  it('Current bookings from police custody suite are not processable', () => {
```
andrewrlee, line 169: no longer handled specifically

server/views/partials/expectedArrivalCardMacro.njk
```diff
@@ -50,9 +50,10 @@
     <div class='app-card__content'>
         <div class="app-card app-card--with-aside">
             <aside class="app-card__aside">
+                {% set hasPhoto = arrival.matchType in ['SINGLE_MATCH', 'COURT_RETURN'] %}
```
andrewrlee, line 53: court return is a different type of single match

mgreen-mojhub (approved):
LGTM

PR #318, opened 2022-09-12
DCS-1839 Speeding up tests

andrewrlee:
Removing type checking during test running reduces memory usage significantly and allows for running in parallel

I've seen time to run reduce from ~2 mins 25 seconds to ~25 seconds

To do this custom type declaration files (`.d.ts`) cannot contain anything that generates actual js code, e.g enums. 
This seems like a bit of an anti-pattern anyway as quite a bit of tooling (including cypress) fails to allow referencing these


petergphillips (approved):
LGTM

PR #321, opened 2022-09-14
DCS-1851 Deriving court returns by using nomis info

andrewrlee:
This matches the logic we use when doing the redirects

### DIFF
--- a/server/services/matchTypeDecorator.ts
+++ b/server/services/matchTypeDecorator.ts
@@ -42,5 +42,5 @@ export class MatchTypeDecorator {
   }
 
   private isCourtReturn = <T extends ArrivalInfo>(item: T): boolean =>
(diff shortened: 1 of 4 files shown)


server/services/matchTypeDecorator.ts
```diff
@@ -42,5 +42,5 @@ export class MatchTypeDecorator {
   }
 
   private isCourtReturn = <T extends ArrivalInfo>(item: T): boolean =>
-    item.isCurrentPrisoner && item.fromLocationType === 'COURT'
+    item.potentialMatches[0].arrivalType === 'FROM_COURT'
```
andrewrlee, line 45: this looks ropey but this is only called if a single potential match and `FROM_COURT` is type checked from the union

louisenorris (approved):
LGTM

PR #324, opened 2022-09-16
DCS-1852 Fixing trace logging in app insights

andrewrlee:


server/utils/azureAppInsights.ts
```diff
@@ -21,9 +21,7 @@ function version(): string {
 }
 
 export function initialiseAppInsights(): void {
-  // Loads .env file contents into | process.env
-  config()
-  if (process.env.APPINSIGHTS_INSTRUMENTATIONKEY) {
+  if (process.env.APPLICATIONINSIGHTS_CONNECTION_STRING) {
```
andrewrlee, line 24: this was using incorrect env var name, don't quite know how it was working previously

server/utils/azureAppInsights.ts
```diff
@@ -1,4 +1,4 @@
-import { config } from 'dotenv'
+import 'dotenv/config'
```
andrewrlee, line 1: this means we don't have to explicitly call `config()`

jobs/eventsJob.ts
```diff
@@ -1,6 +1,5 @@
-/* eslint-disable import/first */
-/* eslint-disable import/order */
-import 'dotenv/config'
```
andrewrlee, line 3: don't need as this is included as part of the line below

louisenorris (approved):
LGTM

PR #325, opened 2022-09-20
DCS-1852 Move dotenv to dev dependencies

andrewrlee:
Moving dotenv to dev dependencies. This removes some code that is only relevant for local development. This code was a bit tricky as it required running at a specific time in the application lifecycle.

There was a weird bug where the app insight initialise() would fail if the env vars weren't present before the app insights import.

This now means the application runs similar locally to how it would run in docker or kubernetes.

Also moving the app insights import so it's very obviously the first thing that happens when the app starts.


mgreen-mojhub (approved):
LGTM

PR #329, opened 2022-09-21
DCS-1857 redirect court flow to feature not available as short term fix for redirect to court flow

louisenorris:
- When searching for an unexpected arrival a user may find a single matched record
- This record could marked as out on a court visit
- When the user tries to continue against this record, rather than route to the confirm arrival steps the service recognises the record as coming back from court and routes the user to the court return booking flow
- This route results in failure 

We have another ticket to properly address the problem described above. This is a short term fix to redirect court returns to feature-not-available page after we “resolve arrival type”, see [image](https://app.diagrams.net/?title=rejig-resolution&client=1#G1WnwY4nO0jY4owlsXXSf_AAXt4H3zF-YI) 

The existing court return flow, after you select a prisoner from the choose prisoner page still exists.


andrewrlee (approved):
LGTM

PR #336, opened 2022-09-28
DCS-1860: 🔧Fix /search-for-existing/record-found showing incorrect results when state is reset

connormaglynn:


server/routes/bookedtoday/arrivals/searchforexisting/singleExistingRecordFoundController.test.ts
```diff
@@ -5,14 +5,26 @@ import { appWithAllRoutes, stubCookies } from '../../../__testutils/appSetup'
...
 let app: Express
 
 beforeEach(() => {
   config.confirmNoIdentifiersEnabled = true
-  app = appWithAllRoutes({ roles: [Role.PRISON_RECEPTION] })
-
+  expectedArrivalsService.getMatchingRecords.mockResolvedValue([potentialMatch])
+  app = appWithAllRoutes({ services: { expectedArrivalsService }, roles: [Role.PRISON_RECEPTION] })
   stubCookies([
+    // FIXME: Look into why removing the newArrival state breaks the tests
```
connormaglynn, line 27: Not entirely sure why removing the new arrival state cookie breaks the tests. Finding it hard to debug as well 🤔

server/views/pages/bookedtoday/arrivals/searchforexisting/singleExistingRecordFound.njk
```diff
@@ -13,12 +13,18 @@
...
             html: 'If you think this is the wrong record, you can <a class="govuk-body govuk-link govuk-link--no-visited-state" data-qa="search-instead" href="/prisoners/' + data.id + '/search-for-existing-record/new">search for a record using different details</a> or <a class="govuk-body govuk-link govuk-link--no-visited-state" data-qa="create-new" href="/prisoners/' + data.id + '/review-per-details/new">create a new record</a>.'
         }) }}
-        {{ govukButton({
-            classes: "govuk-button govuk-!-margin-bottom-6",
-            href: "/prisoners/" + data.id + "/start-confirmation",
-            text: "Continue",
-            attributes: {'data-qa': 'continue'}
-        }) }}
+
+        <form method="post" action={{ "/prisoners/" + data.id + "/search-for-existing-record/record-found" }}>
```
andrewrlee, line 17: I think we can remove the action and it will inherit the url from where its hosted from?

server/routes/bookedtoday/arrivals/searchforexisting/singleExistingRecordFoundController.ts
```diff
@@ -20,4 +33,25 @@ export default class SingleExistingRecordFoundController {
       })
     }
   }
+
+  public submit(): RequestHandler {
+    return async (req, res) => {
+      const { id } = req.params
+      const searchData = State.searchDetails.get(req)
+      const potentialMatches = await this.expectedArrivalsService.getMatchingRecords(searchData)
```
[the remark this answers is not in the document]
andrewrlee: it shouldn't be optional, by "passing the prison number"  did you mean rendering a hidden param and submitting as part of the form or including it as part of the url? 

Either way I think this is fine

andrewrlee (approved):
LGTM

server/views/pages/bookedtoday/arrivals/searchforexisting/singleExistingRecordFound.njk
connormaglynn: Did try and do a quick google and couldn't find anything very opinionated. 

In this case, I like the specificity since this page could be loaded from any route (technically), but it will always POST data to one defined resource.

server/routes/bookedtoday/arrivals/searchforexisting/singleExistingRecordFoundController.ts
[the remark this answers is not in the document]
connormaglynn: Yeah that was the plan 😁 Thought using the prisonNumber would be more accurate rather than researching on pnc, dob and names since potentially they could bring back different results if they've been modified in the mean time. But I suppose we're talking about edge edge cases here 😓 

Looks like it's optional on the [API side](https://github.com/ministryofjustice/hmpps-welcome-people-into-prison-api/blob/e8d8ea9f2769ff3a150759e366837d0f3db0d713/src/main/kotlin/uk/gov/justice/digital/hmpps/welcometoprison/model/arrivals/Arrival.kt#L66), so maybe that needs updating? 🤔

PR #338, opened 2022-09-30
DCS-1873 Bumping version of node

andrewrlee:


package.json
```diff
@@ -22,7 +22,7 @@
     "lint": "eslint . --cache --max-warnings 0 && npm run typecheck",
     "typecheck": "tsc && tsc -p integration_tests",
     "test": "jest",
-    "test:ci": "jest --runInBand",
+    "test:ci": "jest --maxWorkers=2",
```
andrewrlee, line 25: new version of jest fixes a hanging issue - still fails with default runners but this seems a bit faster than runInBand on circle

louisenorris (approved):
LGTM

PR #340, opened 2022-10-06
DCS-1859 check for search details cookie on review details page

louisenorris:
If there is a record with no identifiers, and a user makes changes on the search for existing page (e.g. adds an identifier for an existing prisoner and changes the DoB). When they click search they should see the single match page with the search details on the left and the found record on the right.

If they choose to create a new record then they will see the details of the original arrival (with original DoB) on the review details page rather than the searched DoB.

This change will prioritise looking at the search details before loading the current arrival so that a user will end up seeing these on the review details page if present.


andrewrlee (approved):
LGTM

PR #341, opened 2022-10-10
DCS-1825: 🔥 Remove old ingress ⬆️ Update dependencies

connormaglynn:


helm_deploy/hmpps-welcome-people-into-prison-ui/values.yaml
```diff
@@ -14,7 +14,7 @@ generic-service:
     host: app-hostname.local    # override per environment
     tlsSecretName: hmpps-welcome-to-prison-cert
     v1_2_enabled: true
-    v0_47_enabled: true
+    v0_47_enabled: false
```
connormaglynn, line 17: have to set to false instead of removing it since the [default is set to true](https://github.com/ministryofjustice/hmpps-helm-charts/blob/ead8a533ef3683c8ffda83b9080723f8ed99dfe1/charts/generic-service/values.yaml#L50)

helm_deploy/values-dev.yaml
```diff
@@ -5,7 +5,8 @@ generic-service:
   replicaCount: 2
 
   ingress:
-    host: welcome-dev.prison.service.justice.gov.uk
+    hosts:
+      - welcome-dev.prison.service.justice.gov.uk
```
connormaglynn, line 9: Not required, but recommended for v2 of generic service https://github.com/ministryofjustice/hmpps-helm-charts/releases/tag/generic-service-2.0.0

helm_deploy/hmpps-welcome-people-into-prison-ui/Chart.yaml
```diff
@@ -5,8 +5,8 @@ name: hmpps-welcome-people-into-prison-ui
 version: 0.2.0
 dependencies:
   - name: generic-service
-    version: 1.5.0
+    version: 2.1.0
     repository: https://ministryofjustice.github.io/hmpps-helm-charts
   - name: generic-prometheus-alerts
-    version: 0.4.1
+    version: 1.1.0
```
connormaglynn, line 11: Removes a cronjob alerts that was deemed unfit, more details here https://github.com/ministryofjustice/hmpps-helm-charts/commit/58c301f701f1daa771a3406dafa7c0491397ed37

.circleci/config.yml
```diff
@@ -1,8 +1,8 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@5
-  slack: circleci/slack@4.4.2
+  hmpps: ministryofjustice/hmpps@6
+  slack: circleci/slack@4.10
```
connormaglynn, line 5: Thought I'd remove the bug fix part, probably always want them to be updated automagically. Assuming CircleCI slack orb is pretty safe and secure.

Wondering whether we should just have it at v4? Save us updating it.

.circleci/config.yml
```diff
@@ -1,8 +1,8 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@5
-  slack: circleci/slack@4.4.2
+  hmpps: ministryofjustice/hmpps@6
```
connormaglynn, line 4: Now depends on the default docker version provided by circle instead of a static one that was provided by the hmpps orb as recommended by circle.

The [breaking change](https://github.com/ministryofjustice/hmpps-circleci-orb/blob/main/release-notes/6.0.0.md) is to remove any **docker_version** in our config, which we don't have any anyways

.circleci/config.yml
connormaglynn: In fact, is the slack orb even used? Can't see any references to **slack/** in the config? 🤔 Maybe it's been abstracted away by the hmpps-orb, and it's a dangling bit of tech-debt?

connormaglynn: Looks like the slack functionality was removed by me when taking out the **check-outdated** job. I'll remove the orb on this PR as well 👍

davidthomas22 (approved):
LGTM

PR #351, opened 2022-12-23
DCS-1967 Adding lock mechanism to prevent double click

andrewrlee:
To address issue with users somehow by passing the standard govuk frontend double click prevention support

### DIFF
--- a/server/data/lockManager.ts
+++ b/server/data/lockManager.ts
@@ -1,0 +1,33 @@
+import type { RedisClient } from './redisClient'
+
+/**
+ * This isn't particularly robust, and there is a failure mode if a replica becomes leader before
+ * the lock has been written back across a majority of nodes.
+ * https://redis.io/docs/manual/patterns/distributed-locks/
+ *
+ * A more robust solution would follow Redlock algorithm but the complexity does not warrant the
+ * extra effort for a short term solution.
+ */
+export default class LockManager {
+  private readonly prefix = 'lock:'
+
+  constructor(private readonly client: RedisClient) {}
+
+  private async ensureConnected() {
+    if (!this.client.isOpen) {
+      await this.client.connect()
+    }
+  }
+
+  public async lock(key: string, durationSeconds: number): Promise<boolean> {
+    await this.ensureConnected()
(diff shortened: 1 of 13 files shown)


server/middleware/doubleClickPreventionMiddleware.ts
```diff
@@ -0,0 +1,32 @@
+import type { RequestHandler } from 'express'
+import crypto from 'crypto'
+
+import { reseller } from 'googleapis/build/src/apis/reseller'
+import logger from '../../logger'
+import LockManager from '../data/lockManager'
+
+const LOCK_ID_KEY = 'lockId'
+
```
andrewrlee, line 9: I had implemented an unlock also but it added a load of complexity and the lock is at the page view level rather than locking an entire arrival so it didnt seem to warrant the complexity - a page refresh would generate a new key so allow another lock to be created

server/data/lockManager.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (!this.client.isOpen) {
+      await this.client.connect()
+    }
+  }
+
+  public async lock(key: string, durationSeconds: number): Promise<boolean> {
+    await this.ensureConnected()
+    const result = await this.client.set(`${this.prefix}${key}`, 'LOCKED', { NX: true, EX: durationSeconds })
+    return Boolean(result)
+  }
```
andrewrlee, line 26: the NX ensures that a value can not be set if it already exists.

server/routes/bookedtoday/arrivals/confirmArrival/index.ts
```diff
@@ -58,12 +62,14 @@ export default function routes(services: Services): Router {
       '/prisoners/:id/check-answers',
       checkNewArrivalPresent,
       redirectIfDisabledMiddleware(config.confirmEnabled),
+      addLockId,
       checkAnswersController.view()
```
andrewrlee, line 66: the `addLockId` on the `get` generates a new lock identifier/key on every render/reload.

the lock is embedded as a hidden input and submitted to the POST controller. 
Double clicks will send the same lock identifier multiple times.    

The `obtainLock` middleware will fail for subsequent requests passing the same id and redirect to the arrival list page.

mgreen-mojhub (approved):
LGTM

PR #356, opened 2023-01-16
DCS-1983 create new record bug fix

louisenorris:
Bug fix to populate new arrival cookie state from search details in search for existing journey. When a user cannot find an arrival match and clicks the create a new record link on the single match or multiple match page: 
Screenshot 2023-01-16 at 18 35 07

This link redirects to `/review-per-details/new` which clears out pre-existing new arrival state, however this does need to be repopulated from search details as when a user continues on from the review page the prison number is deconstructed from the new arrival state as the `start-confirmation` controller redirects to the relevant new arrival confirmation journey.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,7 @@ jobs:
       - checkout
       - run:
           name: Update npm
-          command: 'sudo npm install -g npm@latest'
+          command: 'sudo npm install -g npm@9.2.0'
       - restore_cache:
           key: dependency-cache-{{ checksum "package-lock.json" }}
       - run:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - checkout
       - run:
           name: Update npm
-          command: 'sudo npm install -g npm@latest'
+          command: 'sudo npm install -g npm@9.2.0'
```
louisenorris, line 28: Pin npm to 9.2.0 because of the issue https://github.com/npm/cli/issues/6051

GurnankCheema (approved):
LGTM

PR #359, opened 2023-01-18
DCS-2001 change style of DPS link

GurnankCheema:
Changed the style of the DPS back link in the service unavailable page to 

Screenshot 2023-01-18 at 16 31 18

This change is on top of Michal's change that sets the SERVICE_IS_UNAVAILABLE to true for prod


andrewrlee (approved):
LGTM

PR #366, opened 2023-01-25
DCS-1932: ✨Add breadcrumbs to `New Arrival Summary Page` Part 3

connormaglynn:


server/views/partials/layout.njk
```diff
@@ -14,89 +14,96 @@
...
+{% from "./breadCrumb.njk" import breadCrumb %}
 
 {% block head %}
 
-  {% if googleAnalyticsId %}
-    <!-- Global site tag (gtag.js) - Google Analytics -->
-    <script async src="{{'https://www.googletagmanager.com/gtag/js?id=' + googleAnalyticsId}}"></script>
-    <script src="/assets/js/googleAnalytics.js" id='google-analytics' data-analytics={{googleAnalyticsId}}></script>
-  {% endif %}
+    {% if googleAnalyticsId %}
```
connormaglynn, line 21: Seems like 4 indents are our standard for `.njk` files? So thought I'd leave the auto-formatting changes in 💅

server/views/partials/layout.njk
```diff
@@ -14,89 +14,96 @@
...
+            {{ govukBackLink({
+                text: "Back",
+                classes: "backLinkUrl",
+                attributes: {'data-qa': 'back-link'},
+                href: "#"
+            }) }}
+        {% endif %}
 
-{% endblock %}
+        {% if breadCrumbs.enabled === true %}
```
connormaglynn, line 91: I did try just `{% if breadcrumbs %}` so we wouldn't need the flag. The flag seemed to work better though.

Suppose in the future, we could also infer the backlink is disabled if breadcrumbs are enabled. But thought I'd leave that for now 👀

andrewrlee (approved):
LGTM

PR #374, opened 2023-02-02
DCS-2033 url re-name

louisenorris:


server/routes/bookedtoday/summaryMoveOnlyController.ts
```diff
@@ -1,14 +1,14 @@
 import type { RequestHandler } from 'express'
 import { ExpectedArrivalsService } from '../../services'
 
-export default class SummaryWithoutRecordController {
+export default class summaryMoveOnlyController {
```
andrewrlee, line 4: Should be upper case

louisenorris: sorry - fixed in latest commit

andrewrlee (approved):
LGTM!

PR #375, opened 2023-02-02
DCS-2026 add body scan summary box and count to post arrival summary page

louisenorris:
Body scan summary box as below added to the post arrival summary page:

Screenshot 2023-02-02 at 15 41 28

### DIFF
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -99,7 +99,7 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   njkEnv.addGlobal('dpsUrl', config.dpsUrl)
   njkEnv.addGlobal('temporaryAbsencesEnabled', config.temporaryAbsencesEnabled)
   njkEnv.addGlobal('supportingMultitransactionsEnabled', config.supportingMultitransactionsEnabled)
-  njkEnv.addGlobal('generateCurrentYear', generateCurrentYear())
+  njkEnv.addGlobal('generateCurrentYear', generateCurrentYear)
 
   return njkEnv
 }
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -57,10 +57,6 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
     return calculateAge(value)
   })
 
-  njkEnv.addFilter('generateCurrentYear', string => {
-    return string ? generateCurrentYear() : null
-  })
-
   njkEnv.addFilter('toOptions', (array, id: string, valueKey, textKey) => {
     return array.map((item: Record<string, string>, index: number) => ({
       value: item[valueKey],
@@ -103,6 +99,7 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   njkEnv.addGlobal('dpsUrl', config.dpsUrl)
   njkEnv.addGlobal('temporaryAbsencesEnabled', config.temporaryAbsencesEnabled)
   njkEnv.addGlobal('supportingMultitransactionsEnabled', config.supportingMultitransactionsEnabled)
+  njkEnv.addGlobal('generateCurrentYear', generateCurrentYear())
 
   return njkEnv
 }
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -4,7 +4,7 @@ import moment from 'moment'
 import express from 'express'
 import * as pathModule from 'path'
 import config from '../config'
-import { calculateAge } from './utils'
+import { calculateAge, generateCurrentYear } from './utils'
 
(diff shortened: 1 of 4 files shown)


server/views/pages/recentArrivals/recentArrivalsSummary.njk
```diff
@@ -70,6 +70,17 @@
                             }
                         ]
                     }) }}
+
+                    {{ summaryCard({
+                        classes: "govuk-!-margin-bottom-6",
+                        titleText: "X-ray body scan",
+                        rows: [
+                            {
+                                key: {  text: "Total number of scans", caption: "Since 1 January " + "current year" | generateCurrentYear },
```
louisenorris, line 79: Not 100% sure if generating the current year is best done via a nunjuck filter (does feel strange using a filter on the `"current year"` string above) but think also better than generating via a service method and passing to the controller? Open to thoughts on this.

davidthomas22: It appears to be a nice way to implement this and make it accessible across the nunjucks templates

davidthomas22 (commented):
LGTM

server/views/pages/recentArrivals/recentArrivalsSummary.njk
andrewrlee: I think generating the year in nunjucks is nice.

it might make more sense for you to add it as a function to the nunjucks [globals](https://github.com/ministryofjustice/hmpps-welcome-people-into-prison-ui/blob/72229af6cb1cc4656f450d98ce787c01fe80fe0a/server/utils/nunjucksSetup.ts#L95)

Then you can still call it to generate it dynamically but without having to pass a filtered string

server/utils/utils.test.ts
```diff
@@ -490,3 +491,20 @@ describe('calculateAge', () => {
...
+describe('generateCurrentYear', () => {
+  it('returns current year', () => {
+    const currentYear = moment().year()
+    expect(generateCurrentYear()).toBe(currentYear)
+  })
+
+  it('does not return previous year', () => {
+    const previousYear = moment().subtract(1, 'years').year()
+    expect(generateCurrentYear()).not.toBe(previousYear)
+  })
```
andrewrlee, line 504: When your function is always returning a (relatively) constant value it's not really worth testing whether its value is other things once you've confirmed what that value is. 

Injecting the concept of "now" might be better as this is just a form of double accounting - you're asserting the test behaves how you have implemented it.

A basic test might be 
expect(generateCurrentYear(2020-04-05 00:01:20)).toBe(2020) 

Other possible interesting test cases might include? 
Now is the very start of the year
Now is the end of the year

andrewrlee: (if you do inject the concept of now, you might want have a default param of something that provide now, as passing now in via nunjucks might be faffy)

louisenorris: Cool, refactored this in the latest commit

server/views/pages/recentArrivals/recentArrivalsSummary.njk
louisenorris: Thank you - added to globals as suggested in latest commit

GurnankCheema (dismissed):
LGTM

server/utils/utils.ts
```diff
@@ -137,6 +137,6 @@ export const calculateAge = (dateOfBirth: string) => {
   return moment().diff(dateOfBirth, 'years')
 }
 
-export const generateCurrentYear = () => {
-  return moment().year()
+export const generateCurrentYear = (now: moment.Moment = moment()) => {
```
andrewrlee, line 140: does this moment get created at start up or on every method call? 
You'd want it to be every method call, so maybe a function would be better `nowGetter = () => moment`

where the impl of generateCurrentYear would be: 
nowGetter().year()

server/utils/nunjucksSetup.ts
```diff
@@ -103,6 +99,7 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   njkEnv.addGlobal('dpsUrl', config.dpsUrl)
   njkEnv.addGlobal('temporaryAbsencesEnabled', config.temporaryAbsencesEnabled)
   njkEnv.addGlobal('supportingMultitransactionsEnabled', config.supportingMultitransactionsEnabled)
+  njkEnv.addGlobal('generateCurrentYear', generateCurrentYear())
```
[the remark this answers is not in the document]
louisenorris: eek thanks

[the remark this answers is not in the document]
louisenorris: Think this is now sorted in latest commit

server/utils/utils.ts
louisenorris: Think this is now sorted in latest commit

GurnankCheema (approved):
LGTM

PR #384, opened 2023-02-09
DCS-2008 add over body scan limit warning to pre arrival summery page

louisenorris:
Added red over body scan limit warning to pre arrival summary page as below. This only displays when a person has reached 116+ body scans:

Screenshot 2023-02-09 at 12 11 46


GurnankCheema (approved):
LGTM

PR #385, opened 2023-02-09
DCS-2009 add over body scan limit warning to post arrival summary page

louisenorris:
Added red over body scan limit warning to post arrival summary page as below. This only displays when a person has reached 116+ body scans:
Screenshot 2023-02-09 at 13 16 00


GurnankCheema (approved):
LGTM

PR #389, opened 2023-02-28
DCS-1799 integration tests can stub multiple user roles

GurnankCheema:
need to add multiple roles as the dpslink from pre-arrival prsioner summary has a button to the DPS profile and the button should show only if user has the two required roles

### DIFF
--- a/integration_tests/mockApis/auth.ts
+++ b/integration_tests/mockApis/auth.ts
@@ -5,12 +5,12 @@ import { getRequests, stubFor } from './wiremock'
 import tokenVerification from './tokenVerification'
 import Role from '../../server/authentication/role'
 
-const createToken = (role: Role) => {
+const createToken = (role: Role[]) => {
   const payload = {
     user_name: 'USER1',
     scope: ['read'],
     auth_source: 'nomis',
-    authorities: [role || ''],
+    authorities: role,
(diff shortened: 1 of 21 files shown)


integration_tests/mockApis/auth.ts
```diff
@@ -5,12 +5,12 @@ import { getRequests, stubFor } from './wiremock'
 import tokenVerification from './tokenVerification'
 import Role from '../../server/authentication/role'
 
-const createToken = (role: Role) => {
+const createToken = (role: Role[]) => {
```
andrewrlee, line 8: `roles` would probably be a better name

integration_tests/mockApis/auth.ts
```diff
@@ -5,12 +5,12 @@ import { getRequests, stubFor } from './wiremock'
...
   const payload = {
     user_name: 'USER1',
     scope: ['read'],
     auth_source: 'nomis',
-    authorities: [role || ''],
+    authorities: role,
```
GurnankCheema, line 13: didn't think there is there need for || ''

andrewrlee (dismissed):
LGTM

integration_tests/mockApis/auth.ts
GurnankCheema: pluralised variable. Left type singular

andrewrlee (approved):
LGTM

PR #391, opened 2023-03-01
Dcs 1799 wpip UI add links from arrivals page to person summary

GurnankCheema:
Feature: when user clicks on a prisoner's name in the arrival list then one of two possible summary pages are displayed before the journey resumes to the original result set page. 

court returns, single matches  -> summaryWithRecord
multi-match, no-match, no ids ->  summaryMoveOnly

### DIFF
--- a/server/routes/bookedtoday/summaryWithRecordController.test.ts
+++ b/server/routes/bookedtoday/summaryWithRecordController.test.ts
@@ -108,7 +108,7 @@ describe('GET /prisoner/:id/summary-with-record', () => {
         .expect('Content-Type', 'text/html; charset=utf-8')
         .expect(res => {
           const $ = cheerio.load(res.text)
-          expect($('[data-qa=dps-link]').length).toBe(1)
+          expect($('[data-qa=prisoner-profile]').length).toBe(1)
         })
     })
     it('should not be displayed without Prison Reception role', () => {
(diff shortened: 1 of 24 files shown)


server/routes/bookedtoday/summaryWithRecordController.test.ts
```diff
@@ -139,8 +139,62 @@ describe('GET /prisoner/:id/summary-with-record', () => {
         .expect('Content-Type', 'text/html; charset=utf-8')
         .expect(res => {
           const $ = cheerio.load(res.text)
-          expect($('[data-qa=dps-link]').length).toBe(0)
+          expect($('[data-qa=prisoner-profile]').length).toBe(0)
         })
     })
+
```
GurnankCheema, line 145: tests added here instead of in e2e as faster to run

andrewrlee (approved):
lgtm

server/views/partials/expectedArrivalCardMacro.njk
```diff
@@ -60,7 +60,7 @@
                     <h2 class="app-card__title" data-qa="{{ moveType }}-title-{{ loopIndex }}">
                         {% if user.isReceptionUser %}
```
andrewrlee, line 61: need to check whether this can be removed - I think all staff should be able to see the summary?
But not confirm

PR #394, opened 2023-03-15
DCS-1799 add feature flag preventing prisoner summary displaying in prod

GurnankCheema:
for new arrivals, we only want to display the prisoner summary pages in Dev and Preprod.
So have feature-flagged the prisoner name link to not call /summary for prod.
Not going through the 'summary journey' is what the application currently does in prod.


andrewrlee (approved):
LGTM

PR #396, opened 2023-03-20
Dcs 2072 WPIP UI breadcrumbs

GurnankCheema:
Added breadcrumbs but switched to them back link via feature-flag for two pages in pre and prod

DEV
Home -> breadcrumb
People booked to arrive today  -> breadcrumb
Temporary Absences  -> breadcrumb

PRE & PROD
Home -> breadcrumb
People booked to arrive today  -> backlink
Temporary Absences  -> backlink


server/routes/temporaryabsences/checkTemporaryAbsenceController.test.ts
```diff
@@ -13,12 +13,16 @@ let app: Express
 const temporaryAbsencesService = createMockTemporaryAbsencesService()
 const raiseAnalyticsEvent = jest.fn() as RaiseAnalyticsEvent
 
-beforeEach(() => {
+const createAppWithAllRoutes = () => {
```
GurnankCheema, line 16: to create the app in each test after setting different config.showBreadCrumb values

server/views/partials/layout.njk
```diff
@@ -78,21 +78,27 @@
...
-                classes: "backLinkUrl",
-                attributes: {'data-qa': 'back-link'},
-                href: "#"
-            }) }}
-        {% endif %}
-
-        {% if breadCrumbs.enabled === true %}
-            {{ breadCrumb(breadCrumbs.pageTitle, breadCrumbs.data) }}
-        {% endif %}
+{% block beforeContent %}
```
GurnankCheema, line 81: will eventually get rid of this confusing !== false logic across the app

server/views/pages/bookedtoday/choosePrisoner.njk
```diff
@@ -1,6 +1,18 @@
 {% extends "../../partials/layout.njk" %}
 {% from  "../../partials/expectedArrivalCardMacro.njk" import expectedArrivalCard, transferCard  %}
 {% set pageTitle = "People booked to arrive today" if supportingMultitransactionsEnabled == true else "Select a person to add to the establishment roll" %}
+{% set hasDpsLink = false %}
+
+{% if showBreadCrumb %}
```
GurnankCheema, line 6: showBreadCrumb is a temp config arg which we'll remove later on

server/routes/bookedtoday/choosePrisonerController.test.ts
```diff
@@ -25,6 +25,36 @@ describe('GET /confirm-arrival/choose-prisoner', () => {
     expectedArrivalsService.getArrivalsForToday.mockResolvedValue(new Map())
   })
 
+  it('should only display back navigation odf the breadcrumb type', () => {
```
andrewrlee, line 28: minor typo

server/routes/temporaryabsences/checkTemporaryAbsenceController.test.ts
```diff
@@ -36,7 +40,37 @@ describe('GET checkTemporaryAbsence', () => {
     return request(app).get('/prisoners/A1234AA/check-temporary-absence').expect(302).expect('Location', '/autherror')
   })
 
+  it('should only display back navigation of the back link type', () => {
+    config.showBreadCrumb = false
+    createAppWithAllRoutes()
```
andrewrlee, line 45: this is a bit gnarly - the var on app is a bit of a smell that suggests you have some unnecessary coupling - you need to call a method with a side effect or an apparently unrelated variable (`app`) will be invalid. 
 
Would suggest getting rid of `app` variable and returning it from  `createAppWithAllRoutes()` passing and moving breadcrumb setup into there: 

so you can just do something like:

`return request(createApp({breadcrumb: true}))`

where create app looks like: 
```
const createApp = ({ breadcrumb}: {breadcrumb: boolean}) => {
   config.showBreadCrumb = breadcrumb
  return appWithAllRoutes({
    services: { temporaryAbsencesService, raiseAnalyticsEvent },
    roles: [Role.PRISON_RECEPTION],
  })
}

```

server/views/pages/bookedtoday/summaryMoveOnly.njk
```diff
@@ -4,6 +4,8 @@
 
 {% set pageTitle = "Summary for person booked to arrive" %}
 {% set hasBackLink = false %}
+{% set hasDpsLink = false %}
```
andrewrlee, line 7: is this the wrong way round - it seems like most pages don't have the dps back link?
Would it be better to default to not having the link and only set the variable when you do have it?

GurnankCheema: yes that would make more sense

server/routes/temporaryabsences/checkTemporaryAbsenceController.test.ts
GurnankCheema: I did think of doing that at first but then felt it was more readable having a separate variable.

PR #403, opened 2023-04-04
DCS2079 WPIP UI prevent arrival reconfirmation

GurnankCheema:
When a user attempts to confirm an arrival, they will be prevented from backtracking to any preceding page in the arrival journey for a period of 30 seconds. The app saves the basmId in Redis when a Post is sent to check-answers. 

Backtracking prevention middleware is triggered when a Get call is made to any page in the arrival journey. The middleware checks whether the basmId exists in the redis db. If it does, a redirect to /duplicate-booking-prevention kicks in. 

The integration tests do not test the backtracking prevention as the controller tests are adequate. However the integration tests were modified to clear the redis db before each test run. Otherwise many would fail as we use the same test data for many tests

The controller tests are numerous but very similar. The test modules have a single new test in which the backtrack prevention lock is true. The previous tests still pass with the backtrack lock equal to  false.


integration_tests/tsconfig.json
```diff
@@ -1,10 +1,11 @@
 {
   "compilerOptions": {
-    "target": "es5",
```
GurnankCheema, line 3: needed as failing linting

server/data/lockManager.ts
```diff
@@ -19,9 +19,15 @@ export default class LockManager {
     }
   }
 
-  public async lock(key: string, durationSeconds: number): Promise<boolean> {
+  public async lock(moveId: string, durationSeconds: number): Promise<boolean> {
     await this.ensureConnected()
-    const result = await this.client.set(`${this.prefix}${key}`, 'LOCKED', { NX: true, EX: durationSeconds })
+    const result = await this.client.set(`${this.prefix}${moveId}`, 'LOCKED', { NX: true, EX: durationSeconds })
```
GurnankCheema, line 24: setting a lock

integration_tests/integration/bookedtoday/arrivals/autoMatchingRecords/hasMultipleMatches.cy.ts
```diff
@@ -41,6 +41,7 @@ const arrival = expectedArrivals.arrival({
 
 context('Arrival matches multiple records', () => {
   beforeEach(() => {
+    cy.task('resetRedis')
```
GurnankCheema, line 44: this clears the redis state

server/middleware/backTrackPreventionMiddleware.test.ts
```diff
@@ -0,0 +1,50 @@
+import { setLock, isLocked } from './backTrackPreventionMiddleware'
```
GurnankCheema, line 1: renamed file name

server/routes/bookedtoday/arrivals/autoMatchingRecords/multipleMatchingRecordsFoundController.test.ts
```diff
@@ -57,6 +60,19 @@ describe('possible records found', () => {
           expect($('h1').text()).toContain('Possible existing records have been found')
         })
     })
+    it('should redirect to /duplicate-booking-prevention if arrival already confirmed', () => {
+      lockManager.getLockStatus.mockResolvedValue(true)
```
GurnankCheema, line 64: mock the dependency on which the backtrackPreventionMiddleware relies on

server/views/pages/bookedtoday/arrivals/preventDuplicateBooking.njk
```diff
@@ -0,0 +1,21 @@
+{% extends "../../../partials/layout.njk" %}
```
GurnankCheema, line 1: the page that will be displayed if user travels back after attempting book in

server/data/lockManager.ts
```diff
@@ -19,9 +19,15 @@ export default class LockManager {
...
+    return Boolean(result)
+  }
+
+  public async getLockStatus(moveId: string): Promise<boolean> {
```
andrewrlee, line 28: As discussed, wouldn't this be better named `isLocked`? 

getLockStatus and returning a boolean doesn't really make sense, it doesn't tell you what the boolean represents.

If you were to guess that the boolean represent whether a lock is present or not (arguably a big leap!), how would you know whether true represents locked or unlocked? - there is no context to help you know

cypress.config.ts
```diff
@@ -25,6 +26,8 @@ export default defineConfig({
       on('task', {
         reset: resetStubs,
 
+        resetRedis: resetRedisDb,
```
andrewrlee, line 29: I'd have been tempted to link this to the `reset` task so you don't need to remember to add it in as a separate call for each test, like we do for [use of force](https://github.com/ministryofjustice/use-of-force/blob/main/cypress.config.ts#L32)

server/routes/bookedtoday/arrivals/autoMatchingRecords/index.ts
```diff
@@ -10,15 +10,18 @@ import validate from '../../../../middleware/validationMiddleware'
...
   return Routes.forRole(Role.PRISON_RECEPTION)
-    .get('/prisoners/:id/record-found', singleMatchController.view())
-    .get('/prisoners/:id/no-record-found', noMatchController.view())
-    .get('/prisoners/:id/possible-records-found', mulitipleMatchController.view())
+    .get('/prisoners/:id/record-found', checkIsLocked, singleMatchController.view())
+    .get('/prisoners/:id/no-record-found', checkIsLocked, noMatchController.view())
+    .get('/prisoners/:id/possible-records-found', checkIsLocked, mulitipleMatchController.view())
     .post(
       '/prisoners/:id/possible-records-found',
       validate(MatchedRecordSelectionValidation),
```
andrewrlee, line 27: would it make sense to add it to the POST too?

server/routes/bookedtoday/arrivals/courtreturns/confirmCourtReturnAddedToRollController.test.ts
```diff
@@ -36,6 +40,19 @@ describe('confirmCourtReturnAddedToRollController', () => {
         .expect('Location', '/autherror')
     })
 
+    it('should redirect to /duplicate-booking-prevention if arrival already confirmed', () => {
+      lockManager.getLockStatus.mockResolvedValue(true)
+
+      app = appWithAllRoutes({
```
andrewrlee, line 46: shouldn't need to recreate the app here? you have a reference to the lock manager that it's using any way?

andrewrlee: this applies to the other tests too

server/routes/bookedtoday/arrivals/autoMatchingRecords/index.ts
GurnankCheema: I did think about post but didn't think it was needed. User would need to be on the page to do the post.

server/data/lockManager.ts
GurnankCheema: isLocked was used in the middleware function. I decided to have a diff name within the lock manager module because it makes searching for code a bit easier if things aren't named the same. 
But small matter so have changed to isLocked

server/routes/bookedtoday/arrivals/courtreturns/confirmCourtReturnAddedToRollController.test.ts
GurnankCheema: agreed

cypress.config.ts
GurnankCheema: agreed

server/routes/bookedtoday/arrivals/autoMatchingRecords/index.ts
andrewrlee: The lock could be created whilst they are on the page before clicking submit, so would need to protect this if the POST was doing anything significant side effects/writing data/making api calls etc... 

I think it's fine in this case - the main POST is protected and the other POSTs only modify session state

server/routes/bookedtoday/arrivals/autoMatchingRecords/noMatchingRecordsFoundController.test.ts
```diff
@@ -32,6 +35,19 @@ describe('GET /view', () => {
     return request(app).get('/prisoners/12345-67890/no-record-found').expect(302).expect('Location', '/autherror')
   })
 
+  it('should redirect to /duplicate-booking-prevention if arrival already confirmed', () => {
+    lockManager.isLocked.mockResolvedValue(true)
+
+    app = appWithAllRoutes({
```
andrewrlee, line 41: unneeded app

server/routes/bookedtoday/arrivals/autoMatchingRecords/singleMatchingRecordFoundController.test.ts
```diff
@@ -34,7 +37,19 @@ describe('GET /view', () => {
     app = appWithAllRoutes({ roles: [] })
     return request(app).get('/prisoners/12345-67890/record-found').expect(302).expect('Location', '/autherror')
   })
+  it('should redirect to /duplicate-booking-prevention if arrival already confirmed', () => {
+    lockManager.isLocked.mockResolvedValue(true)
 
+    app = appWithAllRoutes({
```
andrewrlee, line 43: unneeded app?

server/routes/bookedtoday/arrivals/courtreturns/checkCourtReturnController.test.ts
```diff
@@ -25,7 +27,11 @@ const arrival = {
 const arrivalResponse = createArrivalResponse()
 
 beforeEach(() => {
-  app = appWithAllRoutes({ services: { expectedArrivalsService, raiseAnalyticsEvent }, roles: [Role.PRISON_RECEPTION] })
+  lockManager.isLocked.mockResolvedValue(false)
+  app = appWithAllRoutes({
```
andrewrlee, line 31: unneeded app

andrewrlee (approved):
LGTM

PR #411, opened 2023-05-02
DCS-2230 hide any reference to body scans for Styal prison

GurnankCheema:
This is a a quick turn-around temp fix while we decide whether we want to hide body scan using user's current caseload or whether to use prisoner gender

The code changes in the PR have used current caseload


davidthomas22 (approved):
LGTM

PR #416, opened 2023-06-15
DCS-2145 send username to backend for some calls

GurnankCheema:
The BASM service would like to know which users access its service. This will be achieved by adding the 'X-current-user' header to calls made by wpip-api to BASM. However the user details originate in wpip-ui and hence they need to be passed from wpip-ui to wpip-api and onwards to BASM.
There are 2 specific wpip-api endpoints that need the username:- 

1. /arrivals/{arrivalId}
2. /prisons/{prisonId}/arrivals

So this ticket is to add the username to wpip-ui functions that result in hitting those 2 endpoints.


michalnawrockidj (approved):
LGTM

PR #428, opened 2023-08-18
DCS-2228 remove flags and refactor navigation

GurnankCheema:
The SHOW_BREADCRUMB flags were set to true for all environments. As there is no diff between the 3 envs, I have removed the flag and updated config. 

Refactored each njk so that it explicitly sets either a back link or breadcrumb. In the absence of both, no link will be shown at all.

The gist of which link to show: static pages have breadcrumbs, interaction pages where we add details or make selections have back links.

hence these show breadcrumbs 
/choose-prisoner
/prisoners-returning
 /recent-arrivals/
 /recent-arrivals/UUID/summary
 /summary-move-only
/summary-with-record

all the rest show a back link apart from the conformation pages (containing green blocks)


GurnankCheema:
decided not to refactor this way

PR #435, opened 2023-09-01
MAP-65 correct layout and remove full stops in accessibility statement

GurnankCheema:
corrections made as per content check by Design team


davidthomas22 (approved):
LGTM

PR #444, opened 2023-10-06
MAP-343: Upgrade Helm chart to be compatible with k8s 1.25

jimbali:
Upgrade Helm chart to be compatible with k8s 1.25

https://dsdmoj.atlassian.net/browse/MAP-343


helm_deploy/hmpps-welcome-people-into-prison-ui/Chart.yaml
```diff
@@ -5,7 +5,7 @@ name: hmpps-welcome-people-into-prison-ui
 version: 0.2.0
 dependencies:
   - name: generic-service
-    version: 2.1.0
+    version: 2.6.5
```
jimbali, line 8: Helm diff shows that the only thing this will change is adding the pod anti-affinity

danbenton-mojdt (approved):

PR #451, opened 2023-11-27
MAP-481 Bug - Location link

davidthomas22:
Bug caused by new header/footer API ordering issue when compiling the WPIP application. Switched `layout.njk` compilation order to resolve


GurnankCheema (approved):
LGTM

PR #458, opened 2024-05-08
MAP-1045 dependency update

GurnankCheema:


package.json
```diff
@@ -174,15 +174,12 @@
...
     "nodemon": "^3.0.1",
     "prettier": "^2.8.8",
-    "sass": "^1.62.1",
+    "sass": "^1.77.0",
     "supertest": "^6.3.3",
     "ts-jest": "^29.1.2",
     "typescript": "^4.9.5"
   },
   "overrides": {
-    "cypress": {
```
GurnankCheema, line 183: no need to override anymore as Cypress 13.8.1 already has v4.1.4

marcusaleman (approved):
I see dependency update.
@GurnankCheema  let me know if it builds and your dev test works ok then I am happy.

PR #461, opened 2024-06-18
MAP-1280 Install postgres cert

danbenton-mojdt:


Mjwillis (approved):
LGTM

Mjwillis (changes requested):
Actually is this needed?  Does UI talk to a postgres DB?

PR #462, opened 2024-06-20
Revert "MAP-1280 Install postgres cert"

danbenton-mojdt:
Reverts ministryofjustice/hmpps-welcome-people-into-prison-ui#461


Mjwillis (approved):
LGTM

PR #471, opened 2024-10-28
Map 1688 wpip UI user prison register to get prison data

GurnankCheema:
Get prison name from prison-register rather than welcome-api
Use the prison-register `prison/id/:prisonId `endpoint. 
The only bit of info needed from the object returned from prison-register is the `prisonName`. This is the same as the 'description' previously obtained when calling welcome-api


Mjwillis (approved):
LGTM

PR #477, opened 2024-11-20
MAP-1859 updated to node v22

GurnankCheema:
This PR is only to update node to v22. Because of this, I have used response type of 'any' in /setUpHealthChecks.ts as temporary a workaround until we align all modules with the template project.
Updates for other dependencies will be in another PR

[See PR in template project that also updates to node 22](https://github.com/ministryofjustice/hmpps-template-typescript/pull/474/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ parameters:
 
   node-version:
     type: string
-    default: 20.15-browsers
+    default: 22.10-browsers
 
 jobs:
   build:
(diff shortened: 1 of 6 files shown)


marcusaleman (approved):
LGTM.
nvm install at local is ok @GurnankCheema  ? right

PR #486, opened 2024-12-09
MAP-1858 merge template typescript pr3 redis

marcusaleman:
Need to propagate value to values.yaml


helm_deploy/hmpps-welcome-people-into-prison-ui/values.yaml
```diff
@@ -28,6 +28,7 @@ generic-service:
   # Environment variables to load into the deployment
   env:
     NODE_ENV: "production"
+    REDIS_ENABLED: "true"
```
marcusaleman, line 31: This is the only commit extra from the previous PR as we need this value to be able to deploy into DEV environment
@GurnankCheema @michalnawrockidj

michalnawrockidj (approved):
LGTM

PR #491, opened 2025-01-06
CDPS-1080: Switching to use the frontend components library

brightonsbox:


server/config.ts
```diff
@@ -106,14 +106,6 @@ export default {
...
       agent: DEFAULT_AGENT_CONFIG,
     },
-    frontendComponents: {
-      url: get('COMPONENT_API_URL', 'http://localhost:8082', requiredInProduction),
-      timeout: {
-        response: Number(get('COMPONENT_API_TIMEOUT_SECONDS', 2000)),
-        deadline: Number(get('COMPONENT_API_TIMEOUT_SECONDS', 2000)),
-      },
-      agent: DEFAULT_AGENT_CONFIG,
-    },
```
brightonsbox, line 116: We still need `COMPONENT_API_URL`  providing in the helm config, but the library retrieves that configuration itself without needing to define it here.

server/data/feComponentsClient.ts
brightonsbox, line 1: FE component clients and services can be removed since this is taken care of by the library.

server/middleware/setUpWebSecurity.ts
```diff
@@ -45,13 +45,6 @@ export default function setUpWebSecurity(): Router {
   ]
   const fontSrc = ["'self'"]
 
-  if (config.apis.frontendComponents.url) {
-    scriptSrc.push(config.apis.frontendComponents.url)
-    styleSrc.push(config.apis.frontendComponents.url)
-    imgSrc.push(config.apis.frontendComponents.url)
-    fontSrc.push(config.apis.frontendComponents.url)
-  }
-
```
brightonsbox, line 54: The library handles this for us so we can remove the custom code from the CSP.

server/views/partials/layout.njk
```diff
@@ -86,11 +81,7 @@
 {% endblock %}
 
 {% block header %}
-  {% if feComponents.header %}
     {{ feComponents.header | safe }}
-  {% else %}
-    {% include "./header.njk" %}
-  {% endif %}
```
brightonsbox, line 93: Fallback header and footer is taken care of by the library

server/data/feComponentsClient.ts
brightonsbox, line 1: The clients and services are now provided by the library

GurnankCheema (approved):
LGTM

PR #495, opened 2025-01-20
MAP-2071 reduced the data capture to last 6 months

GurnankCheema:
Job for the Looker  Studio report crashes because it takes too long to create a report. We are currently getting 1 years worth of the data. Reducing that to 6 months to see if that stops the crash

### DIFF
--- a/jobs/eventsRetriever.test.ts
+++ b/jobs/eventsRetriever.test.ts
@@ -20,8 +20,8 @@ describe('test', () => {
       Readable.from('a,b,c\n1,2,3\n4,5,6').pipe(writable)
     })
 
-    const dataStartDate = moment().subtract(1, 'years').startOf('day')
-    const events = await eventsRetriever.retrieveEventsForPastYear()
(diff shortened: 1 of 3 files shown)


jobs/eventsRetriever.test.ts
```diff
@@ -30,7 +30,7 @@ describe('test', () => {
 
     const call = welcomeClient.getEventsCSV.mock.calls[0]
     expect(dataStartDate.isSame(call[1])).toBe(true)
-    expect(call[2]).toBe(365)
+    expect(call[2]).toBe(183)
```
Thource, line 33: Will 183 always be correct? wary that this test may fail next month

Thource: Sorry, just seen that this `183` is hard coded in the function below

GurnankCheema: I used 183 to account for leap years. Essentially to code will look at the period from 6 months back to today and fetch all the wpip moves in that period.  If the report was being requested on 29 Feb the 183rd day will capture it.

PR #723, opened 2026-07-22
MAPA-218: Send the imprisonment status's fixed from location on confirm

Mjwillis:
## Why

Supports the new "Repatriated to this country" admission type added in [hmpps-welcome-people-into-prison-api#360](https://github.com/ministryofjustice/hmpps-welcome-people-into-prison-api/pull/360).

HMP Wandsworth receive British citizens repatriated or extradited back into the UK. In NOMIS they have always recorded these with movement reason "Imprisonment without option of a fine" **and** a point of origin of the agency location "Repatriated To This Country" (`FORGN`).

WPIP could do the movement reason but not the point of origin. `confirmUnexpectedArrival` hardcoded `fromLocationId: undefined`, so every manually confirmed arrival — the path Wandsworth actually use — landed in NOMIS as coming from `OUT`.

The radio itself needs no UI change; the list is rendered straight from `GET /imprisonment-statuses`. This PR is only the plumbing to carry the status's fixed origin through to the confirm call.

## What

- `fromLocationId` added to the `ImprisonmentStatus` type and to `NewArrival` state (type, codec `write`, `read` allowlist and returned object).
- Both status controllers now copy `fromLocationId` from the selected status into state, on the single-reason and multi-reason paths.
- `confirmUnexpectedArrival` sends it instead of always sending nothing.
- `confirmExpectedArrival` uses `data.fromLocationId ?? arrival.fromLocationId`.

**On that fallback:** for an expected arrival BaSM already knows where the person physically travelled from, and that is the more accurate record, so it wins. The status-supplied default only fills the gap. Wandsworth's case is the manual path, which has no BaSM value, so it always gets `FORGN`.

## Testing

`npm test` (728 pass) and `npm run lint` clean. New coverage for: the status controller storing the fixed origin, the unexpected-arrival confirm forwarding it, and both branches of the expected-arrival precedence rule.

Cypress not run locally. Existing specs select radios by code so the extra fixture option is inert for them.

## Notes

- The API PR must merge first — this is inert without it.
- `integration_tests/mockApis/responses/imprisonmentStatuses.ts` has drifted from the API (old `SENT`/`RX`/`JR` codes, and it still calls the transfer entry "Transfer from another establishment"). I added the new entry using the file's existing convention rather than refreshing it, because several cypress specs assert `imprisonmentStatus: 'SENT'` in the confirm payload and a wholesale refresh would break them. Worth a separate ticket.

## Before release

Verify end to end in dev/preprod that a manual arrival with this option produces a NOMIS record showing movement reason "Imprisonment without option of a fine" and point of origin "Repatriated To This Country".

### DIFF
--- a/server/data/__testutils/testObjects.ts
+++ b/server/data/__testutils/testObjects.ts
@@ -238,10 +238,18 @@ export const createImprisonmentStatuses = (): ImprisonmentStatus[] => [
       { description: 'Partly suspended sentence', movementReasonCode: 'P' },
     ],
   },
+  {
+    code: 'repatriation',
+    description: 'Repatriated to this country',
+    imprisonmentStatusCode: 'SENT',
+    fromLocationId: 'FORGN',
+    movementReasons: [{ movementReasonCode: 'I' }],
+  },
 ]
 
 export const statusWithSingleReason = createImprisonmentStatuses().find(s => s.code === 'on-remand')
 export const statusWithManyReasons = createImprisonmentStatuses().find(s => s.code === 'determinate-sentence')
+export const statusWithFromLocation = createImprisonmentStatuses().find(s => s.code === 'repatriation')
 
 export const createMatchCriteria = (): PotentialMatchCriteria => ({
   firstName: 'James',
@@ -278,7 +286,7 @@ export const createNewArrival = ({
   movementReasonCode = 'N',
   prisonNumber = 'A1234AA',
   expected = true,
-} = {}): NewArrival => ({
+}: Partial<NewArrival> = {}): NewArrival => ({
   firstName,
   lastName,
   dateOfBirth,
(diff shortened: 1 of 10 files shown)


server/data/__testutils/testObjects.ts
```diff
@@ -238,10 +238,18 @@ export const createImprisonmentStatuses = (): ImprisonmentStatus[] => [
       { description: 'Partly suspended sentence', movementReasonCode: 'P' },
     ],
   },
+  {
+    code: 'repatriation',
+    description: 'Repatriated to this country',
+    imprisonmentStatusCode: 'SENT',
```
danbenton-mojdt, line 244: Should this be RECEP_DET?

Thource (approved):

